PR #3, opened 2016-01-28
New Test Program Flow Generator Backend

ginty:
### Summary

Previously, each test platform implemented its own test program flow generation API. Although effort had been made to make these look similar across Teradyne and Advantest platforms, the backend implementation was completely different.

This change implements a single user facing API to describe the test flow, to be used regardless of the target platform. This builds an abstract model of the flow in the background, which is only converted to a platform-specific view at final write time.

### Benefits

Flow error checking and optimization used to be done by each platform-specific implementation, which was error-prone, inefficient and resulted in very complicated code that was hard to reason about and debug.

The semantic representation and analysis of the test flow has now been fully offloaded to a separate plugin, called ATP (abstract test program).
This has resulted in much better flow optimization algorithms (e.g. combining adjacent tests with the same execution conditions) and flow logic error checking (e.g. checking for duplicate or missing test ID references).

Furthermore this abstract flow model is ideal for documentation rendering and has therefore completely removed the need for a dedicated documentation interface at application level.

By reducing the responsibilities of the platform-specific code to final rendering, it has allowed for much more advanced code generation to be implemented. For example the Teradyne platforms now fully support nested, AND'd and OR'd test conditions.

### API Changes

The OrigenTesters::Doc (documentation tester) model and interface API has been completely removed. Apps should therefore remove their documentation interface, targets and any other references to OrigenTesters::Doc.

The skip API has been removed. I believe this was generally used in the past to work around the fact that nested enable words were not supported on Teradyne, but now they are. Therefore just replace skip with the underlying condition that applies to it:

~~~ruby
# Before
skip unless_enable: "nvm_bootloader_gate" do
  import "../../components/flsloader"
end

# After
if_enable "nvm_bootloader_gate" do 
  import "../../components/flsloader"
end
~~~

Any platform-specific flow APIs have been removed, for example consider this interface method:

~~~ruby
# Before
def log(message, options = {})
  if tester.j750?
    flow.logprint(message, options)
  elsif tester.v93k?
    flow.print_to_datalog(message, options)
  end
end

# After, though actually this interface method is now redundant and could be removed altogether
def log(message, options = {})
  flow.log(message, options)
end
~~~

The full flow generation API is here: https://github.com/Origen-SDK/origen_testers/blob/atp/lib/origen_testers/flow.rb

The if_any/all_failed/passed conditions no longer mean any/all devices, rather they now mean any or all referenced tests. This is used to implement AND and OR logic:

~~~ruby
func :test1, id: :t1
func :test2, id: :t2

# OR logic
func :test3, if_any_failed: [:t1, :t2]

# AND logic
func :test4, if_all_passed: [:t1, :t2]

# Block forms are available as with all condition APIs
if_any_passed [:t1, :t2] do
  #...
end
~~~

OR'd and AND'd enable words are also supported now:

~~~ruby
# OR
if_enable [:word1, :word2] do
  # ...
end

# AND
if_enable :word1 do
  if_enable :word2 do
    # ...
  end
end
~~~

Finally ORs/ANDs can be applied by using groups, here is an example of how you might implement speed binning logic:

~~~ruby
group "200Mhz", id: :spec200 do
  func :test1_200
  func :test2_200
end

# Bin out to bin 1 if passed at 200 Mhz, here if_passed will apply to all tests in the group
pass 1, if_passed: :spec200

group "100Mhz", id: :spec100 do
  func :test1_100
  func :test2_100
end

# Bin out to bin 2 if passed at 100 Mhz, here if_passed will apply to all tests in the group
pass 2, if_passed: :spec100

# Otherwise we have failed
bin 3
~~~

Finally, you no longer get a platform specific flow line object back when you insert a new test, so custom decorating is no longer possible. If you need a custom line you now have to use render instead, though this should be used very rarely. If you find the need to do this often we should consider adding to the generic API to cover the required use case:

~~~ruby
# I really want to use the Teradyne if any device failed logic, in that case you are on your own
flow.render "         test  test_x    3    4   Fail             if-any flag-true  my_flag      "
~~~

### Known Changes to Generated (Teradyne) Output

The sub-flow from where an ID originated is no longer injected into it.

An enable word applied to > 5 adjacent tests will now generate branch around logic. This is to respect comments in the IG-XL help which say that enable word evaluation is not optimized for test time.

Flag false conditions now generate as 'not flag-true'. This removes previous issues from the tri-state flag logic where 'flag-false' would not evaluate to true in the case where the flag was never previously set to false.

Flow logic is now generated in the case where a test is subject to the pass/fail conditions of more than one other test. There were a few conditions in my app where such logic had been applied by the flow but previously it was being silently ignored(!), now it actually generates the correct logic.


ginty:
@priyavadan, @redxeth, @ThaoHuynhFsl, please review this, to try in your apps add this to your Gemfile:

~~~ruby
gem "origen_testers", git: "https://github.com/Origen-SDK/origen_testers.git", branch: "atp"
~~~

ginty:
Additional feature added - testers:run command to execute and evaluate the flow at the command line.

Whenever you generate a test program now (using any ATE target) the abstract model will be saved to the tmp directory.

You can then run the new 'origen testers:run' command to execute it. Without arguments this will present the flows that you have available to execute within the program.

Selecting a program will then execute it a provide a datalog-style output to the console:

~~~
origen testers:run probe1
~~~

Options can be supplied to the set the runtime conditions, and you should see the output flow adjust accordingly:

~~~
origen testers:run probe1 --job probe1 --enable vt_collection,data_collection --fail test_id1,test_id2
~~~

In time, once the ATE-specific flow generators have been fully debugged, this alone will be a good enough check that your flow logic does what you want it to.

@ThaoHuynhFsl, try it!

ginty:
Added support for references to test IDs outside of the current flow, to do this the ID must be named beginning with "extern_".

The ID will not be appended with a flow signature in that case and the user is responsible for ensuring its uniqueness.

~~~ruby
Flow.create do
  # This ID can only be referenced within this flow
  func :test1, id: :t1
  # This ID can be referenced outside of this flow
  func :test2 id: :extern_t2
end

Flow.create do
  # This will generate an undefined ID error
  func :test3, if_failed: :t1
  # But this will work, since the ID's name indicates that it is an intentional external reference
  func :test4, if_failed: :extern_t2
end
~~~

ginty:
Some updates added to support generation of Use-limit lines on UltraFLEX.

All UltraFLEX test instance objects, include custom test instances, now have the following attributes:

~~~ruby
ins.lo_limit         # Low test limit, assign an array to specify multiple limits
ins.hi_limit         # High test limit, assign an array to specify multiple limits
ins.scale            # Scale for the limits, this will be automatically calculated, 
                     #    but it can be overridden via this attribute
ins.units            # Units for the limits
ins.defer_limits     # Set to true to use the deferred limits calculation feature
~~~
Some examples:

~~~ruby
ins.lo_limit = 5.mV
ins.hi_limit = 7.mV
flow.test(ins)   # =>       Test       blah
                 #          Use-limit              5          7        m

# Defer limits
ins.lo_limit = 5.mV
ins.hi_limit = 7.mV
ins.defer_limits = true
flow.test(ins)   # =>       Test-defer-limits       blah
                 #          Use-limit              5          7        m

# Multiple limits
ins.lo_limit = [5.mV, 5.5.mV]
ins.hi_limit = [7.mV, 8.mV, 10.M]
ins.units = ["V", "V", "Hz"]
flow.test(ins)   # =>       Test       blah
                 #          Use-limit              5          7        m
                 #          Use-limit              5.5        8        m

# Single values can be supplied, e.g. here for units, in which case it will apply to all
ins.lo_limit = [5.mV, 5.5.mV]
ins.hi_limit = [7.mV, 8.mV]
ins.units = "V"
flow.test(ins)   # =>       Test       blah
                 #          Use-limit              5          7        m    V
                 #          Use-limit              5.5        8        m    V

# Partial arrays are also allowed to generate complex cases:
ins.lo_limit = [5.mV, 5.5.mV]
ins.hi_limit = [7.mV, 8.mV, 10.M]
ins.units = ["V", "V", "Hz"]
flow.test(ins)   # =>       Test       blah
flow.test(ins)   # =>       Test       blah
                 #          Use-limit              5          7        m    V
                 #          Use-limit              5.5        8        m    V
                 #          Use-limit                        10        M    Hz
~~~

Additionally an API has been added to manually set runtime flags and then execute tests based on them. The logic generated from this API will be understood by Origen and will be reflected in the resultant flow model (so it will work in the offline runner for example):

~~~ruby
log "Manual flag setting"
test :test1, on_fail: { set_flag: :my_flag }, continue: true
test :test2, if_flag: :my_flag
unless_flag :my_flag do
  test :test3
end
~~~

Finally, an API has been added to give the ability to easily render fully custom flow lines. Its use is discouraged except for extreme corner cases which cannot be handled by the official flow API. A typical use case for IG-XL would be if you need to use the any/all site flag qualifiers.

~~~ruby
line = flow.ultraflex.goto
line.parameter = "MY_LABEL"
#....
line = flow.ultraflex.nop(label: "MY_LABEL")
~~~

PR #5, opened 2016-02-18
Eric, Intial update

ekingham:
Update for Uflex including:
- DSSC support for digcap / digsrc
- Fixed DSSC instruments statement
- Added 'generic' instruments statement for instruments not yet supported (ie upac, DCVI, etc)
- Added implementation of $tester.set_code for uflex
- other minor updates

### DIFF
--- a/approved/ultraflex/subroutines.atp
+++ b/approved/ultraflex/subroutines.atp
@@ -1,20 +1,20 @@
 // ***************************************************************************
 // GENERATED:
-//   Time:    02-Sep-2015 11:39AM
-//   By:      Stephen McGinty
+//   Time:    18-Feb-2016 18:33PM
+//   By:      ra8585
 //   Command: origen g subroutines -t debug_ultraflex.rb
 // ***************************************************************************
 // ENVIRONMENT:
 //   Application
 //     Source:    git@github.com:Origen-SDK/origen_testers.git
-//     Version:   0.5.0
-//     Branch:    master(295ad2cc603)
+//     Version:   0.5.5
+//     Branch:    feature/uflex_update(d3fa0ce00af)
 //   Origen
 //     Source:    https://github.com/Origen-SDK/origen
-//     Version:   0.4.0
+//     Version:   0.5.8
 //   Plugins
 //     origen_arm_debug:         0.4.3
-//     origen_doc_helpers:       0.2.0
+//     origen_doc_helpers:       0.2.1
 //     origen_jtag:              0.12.0
 //     origen_swd:               0.5.0
 // ***************************************************************************
@@ -25,10 +25,9 @@ compressed = yes;
 instruments = {                                                                                 
                VDDHV:DCVS;                                                                      
                VDDLV:DCVS;                                                                      
-               tdi:digsrc 1:msb:parallel:format=binary:unique_sites:auto_cond_disable;                               
-               tms:digsrc 1:msb:parallel:format=binary:unique_sites:auto_cond_disable;                               
-               tdo:digcap 8:lsb:serial:format=twos_complement:data_type=default:auto_cond_disable:auto_trig_disable:store_stv_disable:store_stv_disable;                               
                mto;                                                                             
+               (tdi,tms):DigSrc 1;                                                              
+               (tdo):DigCap 1;                                                                  
 }                                                                                               
                                                                                                 
 srm_vector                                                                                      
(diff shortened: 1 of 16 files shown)


ginty:
Hi Eric,

Thanks a lot for the quality patch, really glad you finally made it here!

This exposes a very UFLEX/DSSC-specific API, and I am wondering if we could/should try to build a higher level abstraction on top?
i.e. any app-level code that you write using these APIs is not going to re-target well to the V93K for example, but could we express the intention here in a higher level way that we could re-target more easily?

One thing that would be nice to add here is some examples of how this should be used, but I see this being used in the tests:

~~~ruby
$dut.digsrc_overlay(:digsrc_pins => [$dut.pin(:tdi), $dut.pin(:tms)], :subr_name => 'digsrc_overlay_testme32', :define => true, overlay_reg: :testme32)
~~~

I can't actually see where the digsrc_overlay method is defined?

However it seems to be being used to generate a register-mapped overlay, and I'm wondering if an API something like this would work:

~~~ruby
# The register model already supports the concept of associating bits with a named overlay
testme32.overlay("overlay_testme32")

# I think perhaps we need to extend that to store a richer object that also
# contains pin information?
testme32.overlay("overlay_testme32", pins: [$dut.pin(:tdi), $dut.pin(:tms)])

# What's missing is then a standard tester API to generate an overlay vector, something like this:
def overlay(id, *pins)
  # Generate a vector to overlay the given ID on the given pins, i.e. using your new dssc APIs
  # in the case of the UltraFLEX driver
end

# Then anytime you then write the register...
testme32.write!

# We can update the drivers (e.g. JTAG) to use this new standard API...
if bit.has_overlay?
  tester.overlay(bit.overlay_str, bit.overlay_pins)
else
  dut.pin(:tdi).drive!(bit.data)
end
~~~

What do you think, does that sound feasible or I am on the wrong track here?

Also is there anything else that is planned to be done with these dssc APIs that could be similarly abstracted?

Finally, manually setting up the header like this should be eliminated:

~~~ruby
$tester.apply_digsrc_settings($dut.pin(:tdi), $dut.pin(:tms), $dut.digsrc_settings)
~~~

I know you have just followed the conventions that were established by the existing code here, but the original dssc stuff was added before this peer review system was put in place and I would have given this feedback at the time if I had the chance. This seems like a good chance to fix it.

So what should happen is that any time you call one of these dssc methods that take a pin argument, you should store them in an instance variable, @digsrc_pins for example. Then when it comes to rendering the header you can inspect this and generate the necessary header content (I think you have already done this anyway with some other attributes). Its fine to provide an API to allow the app to override the header configuration, but without any specific app-level setting it should generate some sensible default that should work in the majority of cases.

Anyway thanks again for the work so far, hopefully we can work together to add some higher level abstractions that will allow these new features to be used within re-targetable application code.

Cheers,
Stephen

approved/ultraflex/subroutines.atp
```diff
@@ -815,163 +814,94 @@ pipe_minus 1                                                     > tp0
...
-                                                                 > tp0                          0 D X D ;
-((tdi,tms):DigSrc = SEND)                                                                       
-                                                                 > tp0                          0 D X D ;
-((tdi,tms):DigSrc = SEND)                                                                       
-                                                                 > tp0                          0 D X D ;
-((tdi,tms):DigSrc = SEND)                                                                       
-                                                                 > tp0                          0 D X D ;
-((tdi,tms):DigSrc = SEND)                                                                       
+((tdi,tms):DigSrc = Start )                                                                     
+repeat 145                                                       > tp0                          0 D X D ;
```
priyavadan: @ekingham we need to look at the implementation here. The vectors appear different (0 0 X 0 vs 0 D X D) which explains the one missing "Send" on the dig source.

lib/origen_testers/igxl_based_tester/ultraflex.rb
```diff
@@ -498,11 +496,213 @@ def store_next_cycle(*pins)
...
+        dssc(pins, dssc_microcode: 'DigCap = Trig')
+      end
+
+      # Call this method to insert the digcap TRIG microcode on the next vector
+      def digcap_trig_next_cycle(*pins)
+        dssc_next_cycle(pins, dssc_microcode: 'DigCap = Trig')
+      end
+      alias_method :digcap_trig!, :digcap_trig_next_cycle
+
+      # Call this method to insert the digcap STORE microcode on the previous vector
```
ginty: Should this not just be the store method?

Seems like dssc should be the default store method on UltraFLEX instead of STV?

lib/origen_testers/igxl_based_tester/ultraflex.rb
```diff
@@ -498,11 +496,213 @@ def store_next_cycle(*pins)
...
+      def digcap_store(*pins)
+        options = pins.last.is_a?(Hash) ? pins.pop : {}
+        options = {
+          offset:         0,
+          dssc_microcode: 'DigCap = Store'
+        }.merge(options)
+        dssc(pins, options)
+      end
```
ginty: Should this not just be the store method?

Seems like dssc should be the default store method on UltraFLEX instead of STV?

Or more generally, allow the choice of approach to be selected by a tester attribute:

~~~ruby
my_reg.store!   # Uses dssc, assuming we make that the default here

tester.store_method = :stv

my_reg.store!   # Uses STV
~~~

Ultimately I think we want to integrate such an attribute with the tester configuration stuff that @scottchennxp added recently, so that it will automatically selects DSSC as the default capture method whenever the required instrumentation is present.

ekingham:
Hi Stephen –

First thanks to Priyavadan for getting me setup, he was quite patient in working through my initial commit.  I apologize for the big update, I see this flow works better on smaller more standalone changes.

My first goal was just to get a plugin we could use for daily work, as we use DSSC in many mixed signal patterns, while being consistent with the existing implementation (which I later learned is not really being used anywhere just yet).  I agree this looks very uflex specific, one reason I had moved it to Ultraflex.rb for now.  Also most of this does the same thing as $tester.store and $tester.store!, just a different opcode (already changed those to pass in an opcode) and vector character (V for capture just like .store, and D or E for DigSrc).  So I see this method could ultimately serve most/all low level use cases and help simplify the API.

We certainly could drive a higher level of abstraction, although we may need to maintain some of the lower level code just to provide flexibility for DSSC and other Uflex instruments, esp in the short term.

See below for responses in-line.

-Eric

lib/origen_testers/igxl_based_tester/ultraflex.rb
ekingham: Store could also be used as-is, provided the instruments statement enables 2 DigCap options:  'auto_trig_enable' and 'store_stv'.

 `instruments = {
    TDO : digcap 1 : data_type=long : auto_trig_enable : store_stv_enable : receive_data=logic;
  }`

With 'auto_trig_enable', you don't need the 'DigCap = Trig' opcode.  This mode does have some differences in error handling.

With 'store_stv', the 'stv' opcode is used to capture data.  All capture instruments in the pattern must use this mode, so it may not be useful in cases where a pattern uses multiple instruments (say on both tdo and portA).

This is a case where having the instruments statement in a separate header file (pulled in with pre-processor) is helpful - allows to easily change between HRAM captures and DSSC just by re-compiling.

ginty: Controlling via the header options sounds pretty good, though again I think that code should be generated automatically via a tester 'capture_method' attribute or similar.

Either way, everytime I speak to anyone about an UltraFLEX conversion (from 750), they always seem to say stop using STVs. So I think we should change here to make the dssc the default that kicks in when you call reg.store!, but have some way to re-configure the tester to do STVs if you really want to.

ginty:
Hi @ekingham ,

See comments below...

Hi Stephen –

First thanks to Priyavadan for getting me setup, he was quite patient in working through my initial commit.  I apologize for the big update, I see this flow works better on smaller more standalone changes.

_**[@ginty] No problem, seems fine to me**_

My first goal was just to get a plugin we could use for daily work, as we use DSSC in many mixed signal patterns, while being consistent with the existing implementation (which I later learned is not really being used anywhere just yet).

**_[@ginty] Yes, so we have an opportunity to discuss this a bit here, and try and get it right_**

  I agree this looks very uflex specific, one reason I had moved it to Ultraflex.rb for now.  Also most of this does the same thing as $tester.store and $tester.store!, just a different opcode (already changed those to pass in an opcode) and vector character (V for capture just like .store, and D or E for DigSrc).  So I see this method could ultimately serve most/all low level use cases and help simplify the API.

**_[@ginty] Cool_**

We certainly could drive a higher level of abstraction, although we may need to maintain some of the lower level code just to provide flexibility for DSSC and other Uflex instruments, esp in the short term.

_**[@ginty] I think having the lower level/tester-specific APIs is fine. However we should try to map them to the generic APIs where possible too though, i.e. reg.store! should really be using the dssc on UltraFLEX without the user having to get involved with a tester-specific API.**_

See below for responses in-line.
…

Hi Eric,

Thanks a lot for the quality patch, really glad you finally made it here!

This exposes a very UFLEX/DSSC-specific API, and I am wondering if we could/should try to build a higher level abstraction on top?
i.e. any app-level code that you write using these APIs is not going to re-target well to the V93K for example, but could we express the intention here in a higher level way that we could re-target more easily?

One thing that would be nice to add here is some examples of how this should be used, but I see this being used in the tests:

~~~ruby
$dut.digsrc_overlay(:digsrc_pins => [$dut.pin(:tdi), $dut.pin(:tms)], :subr_name => 'digsrc_overlay_testme32', :define => true, overlay_reg: :testme32)
~~~

I can't actually see where the digsrc_overlay method is defined?

[EK] This is in lib/origen_testers/test/dut.rb, part of the regression testcase – would you propose this be moved into the Ultraflex.rb itself?

_**[@ginty] No, that's where I expected it to be, I just got confused when I didn't see it contained within this PR, I guess that just means that you are using the original method and haven't made any changes to it.**_

However it seems to be being used to generate a register-mapped overlay, and I'm wondering if an API something like this would work:

~~~ruby
# The register model already supports the concept of associating bits with a named overlay
testme32.overlay("overlay_testme32")

# I think perhaps we need to extend that to store a richer object that also
# contains pin information?
testme32.overlay("overlay_testme32", pins: [$dut.pin(:tdi), $dut.pin(:tms)])

# What's missing is then a standard tester API to generate an overlay vector, something like this:
def overlay(id, *pins)
  # Generate a vector to overlay the given ID on the given pins, i.e. using your new dssc APIs
  # in the case of the UltraFLEX driver
end

# Then anytime you then write the register...
testme32.write!

# We can update the drivers (e.g. JTAG) to use this new standard API...
if bit.has_overlay?
  tester.overlay(bit.overlay_str, bit.overlay_pins)
else
  dut.pin(:tdi).drive!(bit.data)
end
~~~

What do you think, does that sound feasible or I am on the wrong track here?

[EK] This sounds ok at first glance, I can investigate this some now that I’m more familiar with ‘origen test’.  We may still need some lower level API depending on how the DigSrc is used.  I see some examples which use the DigSrc=Start and DigSrc=Stop opcodes, and others that do not.

Also is there anything else that is planned to be done with these dssc APIs that could be similarly abstracted?

[EK] There may be additional use cases we could abstract, I’d need to research further.  Overlay and store are probably the most often used examples, we can start there.

**_[@ginty] Sounds good, that's all Origen really understands anyway_**

Finally, manually setting up the header like this should be eliminated:

~~~ruby
$tester.apply_digsrc_settings($dut.pin(:tdi), $dut.pin(:tms), $dut.digsrc_settings)
~~~

I know you have just followed the conventions that were established by the existing code here, but the original dssc stuff was added before this peer review system was put in place and I would have given this feedback at the time if I had the chance. This seems like a good chance to fix it.

So what should happen is that any time you call one of these dssc methods that take a pin argument, you should store them in an instance variable, @digsrc_pins for example. Then when it comes to rendering the header you can inspect this and generate the necessary header content (I think you have already done this anyway with some other attributes). Its fine to provide an API to allow the app to override the header configuration, but without any specific app-level setting it should generate some sensible default that should work in the majority of cases.

[EK] The original implementation did this with @digsrc_pins – the pattern first calls a method to define the pins, then a method to define the settings, but this didn’t support multiple instruments on different pins in the same pattern, each with different settings.  See below for an example instruments statement with 2 DigSrc instruments from a DAC pattern.  The above call to .apply_digsrc_settings was updated in order to associate pins with settings and add to the instruments statement, one unique statement each time the method is called.

~~~text
instruments = {
               RDACIN:DigSrc 16;
               TDI:DigSrc 1;
               DAC_AP:UltraCapture;
}
~~~

Pins RDACIN and TDI are separate instruments and can have their own settings for width (only one used here), bit_order, mode, format, data_type, etc.  Defining the pins separately from the settings loses this association.

These settings can be quite extensive, and there are many more instruments we don’t support yet (ie UPAC, DCVI, etc), each with their own unique list of possible configurations from the pattern header.

To get around this in the short term I also had added a method to inject ‘generic’ instruments into the header ($tester.add_instrument_string).  I realize this will lead to uflex-specific patterns ☺ – to remedy this (and the above $tester.apply_digsrc_settings), I’d propose use of a header file (#include) which could contain the instruments statement, this is how other tools deal with this problem.  This header file is a few lines maintained outside of the tool, and would contain 90% of the tester-instrument-dependent code.  

_**[@ginty] I don't think I really like the idea of breaking the relationship between Origen and the header to be honest, though for this kind of thing if you want to add an attribute like tester.generate_pattern_header = false to make it then work the way you like that is fine, I wouldn't like that to be the default though.
Generally the header contains a lot of low level controls which I would prefer to fully automate such that the user doesn't even need to think about it whenever possible.
However, we have definitely been missing a way for the user to also manually inject stuff into the header, and I so I definitely welcome that addition in this PR.
For things like the DCVI where Origen doesn't really understand that, then yes of course we need to provide a way for the application to control this themselves. But for any header configuration detail that can be inferred for the APIs used within the pattern, then we should be automating it, at least to put in a default configuration which should work in most cases.
It seems from my (limited) understanding of how dssc is being used here, that dssc header info should be able to be inferred and injected automatically whenever someone uses the store! API on UltraFLEX.**_

The rest is in the microcode which we can more easily abstract into a higher level API (only a few basic commands typically).  We use this header file method currently to support both J750 and Uflex from the same source (simulation or other toolset).

I’ve also included a method to add a header file #include ($tester.add_pattern_include_files) – if this is used to replace the instruments statement we’d need another option to suppress the instruments from the header (would be taken care of in the user’s header file), and would need to move it down where the instrument statement is.  May need more feedback/discussion on this.

_**[@ginty] That sounds good, yes just need to add tester.generate_pattern_header = false, or something similar in addition to this.**_

Anyway thanks again for the work so far, hopefully we can work together to add some higher level abstractions that will allow these new features to be used within re-targetable application code.

Cheers,
Stephen

ginty:
Closing due to inactivity

PR #6, opened 2016-03-03
Added ATE_Hardware Code and related syntatic sugar code for Origen Core

scottchennxp:
Added ATE_Hardware Code and related syntatic sugar code for Origen Core.  User can now use the following to retrieve tester hardware information for test program generation:

$tester.ate_hardware("HSD-M").ppmu
$tester.ate_hardware("VSM").supply

or

$dut.power_pin_group(:vdd).ate_hardware.supply
$dut.pins(:pinname).ate_hardware.ppmu


ginty:
Hi @scottchennxp,

Looks good.

However, since it is a big update that is taking 10% off our (already low) code coverage number, I think we will need to see some tests for this.

@aaronburg should be able to help with this, or you can let me know.

Also you need to think about adding some documentation for this either to http://origen-sdk.org/testers/ or http://origen-sdk.org. You can delay this if you still think the API is evolving, but once it is considered available we need to think about documentation for it.

Thanks!

scottchennxp:
Added tests for ATEHardware Code releasse.

PR #8, opened 2016-03-16
Support for Edgeset/Timeset sheet generation

aaronburg:
Here is an initial commit to conceptually review the addition of timset and edgeset support - for now focused on UltraFlex


lib/origen_testers/igxl_based_tester/base/generator.rb
```diff
@@ -217,6 +261,38 @@ def patgroups(filename = patgroups_filename)
...
+        # If the sheet does not exist yet it will be created.
+        def edgesets(filename = edgesets_filename)
+          f = filename.to_sym
+          return edgeset_sheets[f] if edgeset_sheets[f]
+          p = platform::Edgesets.new
+          p.filename = f
+          edgeset_sheets[f] = p
+        end
+        alias_method :time_sets, :edgesets
+        alias_method :timing_sets, :edgesets
```
ginty: Should be edge_sets I think

lib/origen_testers/igxl_based_tester/base/edgesets.rb
```diff
@@ -0,0 +1,53 @@
...
+
+        # Equality check to compare full contents of edge object
+        def edges_eql?(edge1, edge2)
+          edge1.d_src == edge2.d_src && edge1.d_fmt == edge2.d_fmt &&
+            edge1.d0_edge == edge2.d0_edge && edge1.d1_edge == edge2.d1_edge &&
+            edge1.d2_edge == edge2.d2_edge && edge1.d3_edge == edge2.d3_edge &&
+            edge1.c_mode == edge2.c_mode &&
+            edge1.c1_edge == edge2.c1_edge && edge1.c2_edge == edge2.c2_edge &&
+            edge1.t_res == edge2.t_res && edge1.clk_per == edge2.clk_per
+        end
```
ginty: Looks like you are going to have an edge object, though it is not included here yet.

If so this would be better implemented by defining an == method in the edge object, e.g.
https://github.com/Origen-SDK/origen/blob/master/lib/origen/users/user.rb#L126

ginty:
Hi @aaronburg, I've added a couple of minor comments as I read through.

Overall though, definitely going in the right direction and pretty much exactly as I would have done it.

aaronburg:
Hi @ginty,
Great to hear and thanks for havinga look.  You're correct about the edge object - I had planned on having both an edge object and something like an edge_collection objects which would be basically a buffet of edges that the user has to draw from for building an edgeset. (hmmm... maybe I should call it edge_buffet instead...  )

lib/origen_testers/igxl_based_tester/base/edgesets.rb
aaronburg: I tried this in the past, but ran into the problem that two edges that may contain the same accessor values still may have differrent object ID's, so the equality fails in this case.  This is how I ended up with the direct accessor comparison...

aaronburg:
Hi @ginty,
I've added a bunch more content on the modelling and sheet creation for Edgesets, Timesets, Levelsets, AC Specsets and DC Specsets.  This constitutes a majority of the additions that I would like to add with this commit.  The items that are not yet checked in include the addition of TimesetBasic sheet generation and Pattern modelling (a container to hold all of the attributes of a pattern pertaining to automating timing, levels, patsets and instance creation).  Let me know what you think.

lib/origen_testers/igxl_based_tester/base/edgesets.rb
ginty: @aaronburg, if you implement the == method it won't care about object ID unless you include it in your comparison. In fact the whole point of doing that is precisely because you still want to be able to consider two different objects equal if the attributes you are about are the same.

ginty:
@aaronburg, just gave a quick scan so far, looks good though.
Would like some test coverage for it before merging it.

lib/origen_testers/igxl_based_tester/base/edgesets.rb
aaronburg: @ginty Unfortunately, I found your statement not to be true regarding the == method as defined in the Object class.  Here is a web snippet I found that supports my finding:
string1 = "abc"
class MyObject
end
object1 = MyObject.new
object2 = object1
object3 = MyObject.new

puts "Object 1 is == to object 2: #{object1 == object2}"
puts "Object 1 is eql? to object 2: #{object1.eql? object2}"
puts "Object 1 is equal? to object 2: #{object1.equal? object2}"
puts "Object 1 is == to object 3: #{object1 == object3}"
puts "Object 1 is eql? to object 3: #{object1.eql? object3}"
The output is:
Object 1 is == to object 2: true
Object 1 is eql? to object 2: true
Object 1 is equal? to object 2: true
Object 1 is == to object 3: false
Object 1 is eql? to object 3: false

I did, however, find an alternative to avoid the ugliness I had before:
edge1.instance_values == edge2.instance_values

ginty: Don't think we are talking about the same thing here @aaronburg:

~~~ruby
class Edge
  attr_accessor :d0, :d1

  def initialize(d0, d1)
    @d0 = d0
    @d1 = d1
  end

  def ==(edge)
    if edge.is_a?(Edge)
      d0 == edge.d0 && d1 == edge.d1
    else
      super
    end
  end
end

e1 = Edge.new(1, 2)
e2 = Edge.new(1, 2)
e3 = Edge.new(1, 3)

e1 == e2   # => true
e1 == e3   # => false
~~~

aaronburg: Thanks for clarifying @ginty I’m with you now…

Will redefine the == for  edge and level objects

aaronburg:
@ginty - I've updated the remaining models and sheet generators for this PR, including examples for interface and sheet generation for regression testing.  Pls let me know if you see any issues with the additions.

PR #11, opened 2016-04-20
Allow v93k header to use lower case for groups and aliases

chrisnappi:
Previously all contents of the header in v93k patterns were forced to uppercase - uppercase is only required for base pin names (the name assigned to a channel).  Aliases and pin groups can have lower case.

### DIFF
--- a/lib/origen_testers/smartest_based_tester/base.rb
+++ b/lib/origen_testers/smartest_based_tester/base.rb
@@ -338,9 +338,18 @@ module OrigenTesters
         }.merge(options)
         pin_list = ordered_pins.map do |p|
           if Origen.app.pin_pattern_order.include?(p.id)
-            p.id.to_s.upcase # specified name overrides pin name
+            # specified name overrides pin name
+            if p.is_a?(Origen::Pins::PinCollection) || p.id != p.name
+              p.id.to_s # groups or aliases can be lower case
+            else
+              p.id.to_s.upcase # pins must be uppercase
+            end
           else
-            p.name.to_s.upcase
+            if p.is_a?(Origen::Pins::PinCollection) || p.id != p.name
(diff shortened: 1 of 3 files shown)


lib/origen_testers/smartest_based_tester/base.rb
```diff
@@ -338,9 +338,18 @@ def pattern_header(options = {})
         }.merge(options)
         pin_list = ordered_pins.map do |p|
           if Origen.app.pin_pattern_order.include?(p.id)
-            p.id.to_s.upcase # specified name overrides pin name
+            # specified name overrides pin name
+            if p.is_a?(Origen::Pins::PinCollection) || p.id != p.name
```
ginty: Can we add a size > 1 qualifier to the pin collection case?

Not sure it would ever be hit in reality, but conceptually a single pin could find its way in here with a pin collection wrapper around it.

chrisnappi: Just to the first part right?  (before the ||).

ginty: Yeah, only if it is a pin collection

lib/origen_testers/smartest_based_tester/base.rb
```diff
@@ -338,9 +338,18 @@ def pattern_header(options = {})
...
+              p.id.to_s # groups or aliases can be lower case
+            else
+              p.id.to_s.upcase # pins must be uppercase
+            end
           else
-            p.name.to_s.upcase
+            if p.is_a?(Origen::Pins::PinCollection) || p.id != p.name
```
ginty: I would add the same change here, after that it looks good to me

chrisnappi:
adding size check

PR #14, opened 2016-05-24
Few updates to UF

priyavadan:
- Match loop support added
- Set code, set msb etc... support added
- Options added to the code which was previously hard coded.
- Test Instance limits for UltraFLEX.
-  Initial checkin, changes here are absolutely required to be merged in the next few weeks.

- **Regression check has not been performed yet but will be done in the next few days.** Hoping that the CI can perform some of the basic regressions.

Special thanks to Eric Kingham for the discussions!

### DIFF
--- a/lib/origen_testers/igxl_based_tester/base.rb
+++ b/lib/origen_testers/igxl_based_tester/base.rb
@@ -717,11 +717,7 @@ module OrigenTesters
         $tester.align_to_last
         # cycle(:microcode => "#{$dut.end_of_pattern_label}:") if $dut.end_of_pattern_label
         if options[:end_in_ka]
-          if $tester.j750?
-            $tester.cycle microcode: 'keep_alive'
-          elsif $tester.ultraflex?
-            $tester.cycle microcode: 'keepalive' # PPKK need to add a new keepalive subroutine (either in global subs or a new keepalive subroutine pattern in Pset sheet)
-          end
+            $tester.cycle microcode: "#{@microcode[:keepalive]}"
         else
           if options[:end_with_halt]
             $tester.cycle microcode: 'halt'
--- a/lib/origen_testers/igxl_based_tester/base.rb
+++ b/lib/origen_testers/igxl_based_tester/base.rb
@@ -718,9 +718,9 @@ module OrigenTesters
         # cycle(:microcode => "#{$dut.end_of_pattern_label}:") if $dut.end_of_pattern_label
         if options[:end_in_ka]
           if $tester.j750?
-          $tester.cycle microcode: 'keep_alive'
+            $tester.cycle microcode: 'keep_alive'
           elsif $tester.ultraflex?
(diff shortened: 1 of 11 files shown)


lib/origen_testers/igxl_based_tester/base.rb
```diff
@@ -717,7 +717,11 @@ def pattern_footer(options = {})
         $tester.align_to_last
         # cycle(:microcode => "#{$dut.end_of_pattern_label}:") if $dut.end_of_pattern_label
         if options[:end_in_ka]
+          if $tester.j750?
```
redxeth: Been a while editing base.rb and doing PRs in github, but I thought the whole point here was to not have have base handle all tester-specifics but have it handled in the child classes j750.rb and ultraflex.rb ?

priyavadan: IMO, The initial implementation assumes that the microcode between J750 and UltraFLEX is the same. the microcode keep_alive is specific for J750, Ultraflex has the microcode keepalilve. So, would we have a pattern_footer method in ultraflex.rb and override this there? Just trying to understand here.

same for below, UF does not have a hi lo limit valid argument.

redxeth: It does not assume that they are the same.  The base is where they are in common, i.e they happen to be the same.  If different is needed then will need to describe them generically in base (say a method or option) and differentiate how those get defined in classes that will inherit from base.  Here the microcode difference could be called @keepalive as used in base where that is defined in j750.rb and ultraflex.rb.  Compare the initialize of those files to see other diffs.

priyavadan: Updates have been pushed and the PR is now updated. Thanks for reviewing!

I am hoping we can have a merge/release in the next couple of days.

ginty:
LGTM :+1:

priyavadan:
Thanks @ginty , just waiting on @redxeth to have a look as well before I merge.

redxeth:
Looks good!  Ship it.

redxeth:
That said-- what level of regression is done here-
does it generate flow sheets and compare them?  Patterns?  How advanced is the model?

Basically am seeing some issues due to ATP updates on my branch, but expect might be able to fix those.  We need to be sure we're running flow file regressions that use as many features as possible.  Is that what coveralls is doing?

priyavadan:
@redxeth coveralls checks for overall documentation coverage (basically comments in the code to describe the code)

The origen test command does check for some patterns, flows, patsets etc for j750, j750 hpt, ultraflex, v93k. 

it essentially is similar to the "examples" command that compares against the approved folder. The two changes in the approved folder in this PR are for UltraFLEX only, so J750/other platforms are still unchanged and a pass confirms that.

priyavadan:
Will merge and make the release later this evening. Thank you all for the quick review!

ginty:
@redxeth, if you do indeed find a regression which was caused by something that was not caught by the OrigenTesters test suite, then we need to update it to include whatever it is.

PR #16, opened 2016-06-14
Allow lower-case name for case of group with only 1 pin

rlaj:


priyavadan:
@rlaj looks good to me.

ginty:
@rlaj I've made you an owner of this gem, release at will

PR #18, opened 2016-06-30
Add J750 PowerSupply_T support and also added pattern alias the the c…

ThaoHuynhFsl:
…urrent apmu_powersupply


priyavadan:
@ThaoHuynhFsl thanks! as discussed let us try to add some coverage to this if possible. I leave on PTO tomorrow evening, so will most likely close this PR and make a new origen_testers release before I leave so you can use it.

ThaoHuynhFsl:
hi @priyavadan :  i just added the test cases for power supply to j750 and also test case to the hi_v (board_pmu), low_v (pin_pmu) that was missing on the current j750 as well.

priyavadan:
@ThaoHuynhFsl Thanks! Unfortunately, coverage seems to have gone down further :). I'll look at after I get back from PTO, will make a origen_testers release for you today before I leave.

PR #21, opened 2016-09-09
93K Comment updates

chrisnappi:
Put header comments at end of line
Move comments before ';' so they pass through to BINL
Updated examples due to above changes


ginty:
I think this should be enabled by default on V93K, can't see a reason why you would ever not want to include inline comments.

lib/origen_testers/smartest_based_tester/base.rb
```diff
@@ -379,18 +379,18 @@ def local_subroutines # :nodoc:
...
         if vec.repeat # > 1
           microcode = "R#{vec.repeat}"
         else
           microcode = vec.microcode ? vec.microcode : ''
         end
         if vec.pin_vals && ($_testers_enable_vector_comments || vector_comments)
-          comment = " # #{vec.number}:#{vec.cycle} #{vec.inline_comment}"
+          comment = " #{vec.comments.join("\cm")} #{vec.number}:#{vec.cycle} #{vec.inline_comment}"
         else
           comment = vec.inline_comment.empty? ? '' : " # #{vec.inline_comment}"
```
ginty: I think this branch should have the same update

PR #22, opened 2016-09-12
93K Comment updates

chrisnappi:
- Add header comment to inline with embedded CR
- Always print cycle/vector comments inline for 93K


config/version.rb
```diff
@@ -1,8 +1,8 @@
 module OrigenTesters
   MAJOR = 0
-  MINOR = 7
-  BUGFIX = 13
-  DEV = nil
+  MINOR = 8
+  BUGFIX = 0
+  DEV = 0
```
ginty: You should manually edit this back to 7.13.0

doc/history
```diff
@@ -1,3 +1,13 @@
+<a class="anchor release_tag" name="v0_8_0_pre0"></a>
+<h2><a href="#v0_8_0_pre0">Tag: v0.8.0.pre0</a></h2>
+
+##### Branch: 'master'
+
+##### by Chris Nappi on 12-Sep-2016 08:48AM
+
+
+Always print inline vector/cycle comments on 93K
+
```
ginty: And manually delete this

ginty:
Hi Chris,

Looks better, but still not quite right. I think we need to put back the original vector comments guard and just update both branches of that if to include these header comments.

The rationale is that we don't really want the vector comments in most of the time, since they produce a wall of diffs any time an additional vector is inserted or something like that.

chrisnappi:
Okay - I misunderstood.  You mean like this:

`        if vec.pin_vals && ($_testers_enable_vector_comments || vector_comments)
          comment = " #{vec.comments.join("\cm")} #{vec.number}:#{vec.cycle} #{vec.inline_comment}"
        else
          comment = " #{vec.comments.join("\cm")}" + vec.inline_comment.empty? ? '' : " # #{vec.inline_comment}"
        end
        "#{microcode.ljust(25)}#{timeset.ljust(27)}#{pin_vals}#{comment};"`

ginty:
Yeah

PR #24, opened 2016-09-12
93K Comment updates

chrisnappi:
Include header comments in-line with embedded 'ctrl-m' so they are visible in gui
Move comments before ';' so they translate into binl


ginty:
Alright, LGTM

ginty:
Merge and release at will btw

PR #25, opened 2016-09-12
Updates to IGXL based testers

priyavadan:
minor updates based on the issues observed in the C55 application.


ginty:
Any thoughts as to why this is required?

Seems like a more general problem with your app. The only way I think it could end up here with tester returning nil, would be if your application is invoking a tester cycle while it is being loaded.

PR #27, opened 2016-10-21
93K touch-ups

chrisnappi:
Unrolled repeats less than 33 rather than 17 - may want to make this selectable from app later.
Cut comments at 250 chars for end-of-line (this is the limit)


ginty:
Looks good to me, remember to release as a minor prod release rather than development

PR #30, opened 2017-01-24
Add test number to testlimits section in tf template

rlaj:


lib/origen_testers/smartest_based_tester/base/test_methods/limits.rb
```diff
@@ -28,13 +30,17 @@ def unit=(val)
 
           def to_s
             if !lo_limit && !hi_limit
-              "\"#{test_name}\"" + ' = "":"NA":"":"NA":"":"":""'
+              if tnum == ""
```
ginty: This if is redundant, "#{nil}" => ""

rlaj: If tnum is not specified by the flow-level decorator, then the extra '0' will not be placed on the testlimit line (i.e. there will be no change to existing users)

ginty: Oh yeah, sorry I missed the 0 at the end.

Good update btw, I've been wondering how test numbers should be handled on 93k.

PR #35, opened 2017-02-28
Add FlagOptimizer processor to the V93K flow finalize method.

rlaj:
For the case where a flow flag was created for on_fail or on_pass, and then immediately used (exclusively) to conditionally run a test(s), for V93K we want to instead place those tests in the if/else clause directly rather than using a flag.  This update creates a V93K-specific processor to transform the AST before rendering the flow.


approved/v93k/flow_control.tf
```diff
@@ -1655,50 +1655,28 @@ test_flow
...
-  }
-  if @T1_BEA7F3B_FAILED == 1 then
-  {
     run_and_branch(erase1_BEA7F3B)
     then
     {
     }
     else
     {
-      stop_bin "", "fail", , bad, noreprobe, red, 12, over_on;
```
ginty: Is this not showing that this bin information has been lost?

lib/origen_testers/smartest_based_tester/base/processors/flag_optimizer.rb
```diff
@@ -0,0 +1,107 @@
...
+              end
+            end
+            results << node_a unless node_a.nil?
+            node.updated(nil, [name] + results)
+          end
+
+          # Given two adjacent nodes, where the first (a) is a test and the second (b)
+          # is a run_flag, determine if (a) conditionally sets the same flag that (b)
+          # uses.  If it does, do a logical replacement, if not, move on quietly.
+          def on_boolean_substitution(node_a, node_b)
```
ginty: How do we handle the case that there is a later test which also references this flag, in that case the flag setting must be preserved, though you could still collapse this node as long as it also keeps the flag setting.

I think we need to make sure we have a test case here for that scenario and I think maybe upfront in this processor we should walk the AST to generate a table which contains the number of occurrences of each flag, then here you can do a simple check to see if count == 1.
In ATP there is a number of examples of such processors which summarize AST data like that.

lib/origen_testers/smartest_based_tester/base/processors/flag_optimizer.rb
```diff
@@ -0,0 +1,107 @@
...
+                node_a = node_b
+              end
+            end
+            results << node_a unless node_a.nil?
+            node.updated(nil, [name] + results)
+          end
+
+          # Given two adjacent nodes, where the first (a) is a test and the second (b)
+          # is a run_flag, determine if (a) conditionally sets the same flag that (b)
+          # uses.  If it does, do a logical replacement, if not, move on quietly.
```
ginty: I think the name of this should be changed, on_boolean_substitution doesn't make sense in this context.
Should be called remove_run_flag or similar.

rlaj: done

lib/origen_testers/smartest_based_tester/base/processors/flag_optimizer.rb
rlaj: done

approved/v93k/flow_control.tf
rlaj: Yes, thanks.  I think the (continue) that is originally with the set-flag is overriding the binning of substituted block from on-flag condition.  Will fix.

approved/v93k/flow_control.tf
```diff
@@ -2581,15 +2494,8 @@ test_flow
   }
   else
   {
-    @MY_FLAG = 1;
```
rlaj: This isn't correct either, marking it for improvement

approved/v93k/flow_control.tf
rlaj: For this example flow:
  log "Test that if_failed works"
  func :read1, id: :t1, bin: 10
  func :erase1, if_failed: :t1, bin: 12

Does the fact that erase1 uses if_failed: :t1, make read1 a continue on fail?  Or does the bin: 10 have priority on the erase1 failure?

ginty: The fact that read1 has a later test dependent on its result overrides the bin, so read1 should be a continue on fail instead of binning out to 10.
The 10 should still stick around if the tester API permits it, e.g. in IG-XL it should still appear in the bin column, but on SMT I think the stop_bin statement just disappears.

rlaj:
I think I covered all the issues mentioned above, please give it another look and let me know if you see any other issues.

approved/v93k/flow_control.tf
rlaj: OK.  Latest update renders the SMT flow accordingly.

approved/v93k/flow_control.tf
rlaj: Corrected in latest commits.

ginty:
@rlaj, seeing as how you are on a roll with this, I don't suppose you fancy making a further update to make tests with an empty branch render as a run_test, rather than a run_and_branch?

That's been annoying me lately.

approved/v93k/flow_control.tf
```diff
@@ -1725,38 +1795,24 @@ test_flow
...
-  if @T6_BEA7F3B_RAN == 1 then
-  {
     run(read0_1_BEA7F3B);
     run(read0_2_BEA7F3B);
   }
   else
   {
+    stop_bin "", "fail", , bad, noreprobe, red, 10, over_on;
+    @T6_BEA7F3B_RAN = 1;
   }
```
ginty: This doesn't seem quite right.
The intention of if_ran is that the parent test should be set to continue on fail and a _RAN flag should be set whether it passed or failed.
The dependent test should therefore run whether the parent test passed or failed, but here it now runs only if the parent test passes and it sets a now redundant flag in the fail case.
The original implementation doesn't look quite right either as the stop_bin should have been removed.

You could just remove the RAN flags from your optimization, checking for auto_generated and name containing _RAN should be good enough.

approved/v93k/flow_control.tf
```diff
@@ -2597,6 +2611,27 @@ test_flow
...
+  run(test22c_BEA7F3B);
+  run(test22d_BEA7F3B);
+  if @AT22_BEA7F3B_FAILED == 1 then
+  {
+    run(test22e_BEA7F3B);
+    run(test22f_BEA7F3B);
+  }
+  else
+  {
+  }
```
ginty: Nice

approved/v93k/flow_control.tf
rlaj: Hadn't considered the if_ran case, will make the updates.

rlaj: Patched the FlagOptimizer processor to ignore the RAN flags.  Created a new processor to remove the set-result node in the fail branch for if_ran conditionals.

rlaj:
Merged.  Will move minor run/branch request to a different PR.

PR #36, opened 2017-03-03
Add empty branch cleaner for V93K flow rendering

rlaj:
This update optimizes away cases where empty then/else branches get created (V93K-specific)


rlaj:
Wasn't quite right, patched and added another test case.

PR #38, opened 2017-03-30
Added support for utility pins for functional test on J750 tester.

coreyeng:
Currently, utility pins are not supported when trying to use a J750 functional test, even though this is a valid thing to do in IGXL. Added support for these by adding them to the functional test hash. Names are kept the same as the other templates (bpmu, ppmu, etc).

For example,
ins.utility_pins_0 = "pin_0"
ins.utility_pins_1 = "pin_1"

will yield the following in the IGXL sheet:
image


ginty:
Hi @coreyeng, thanks for the patch.

Can you run lint on the changed file and then we can merge.

PR #41, opened 2017-06-12
Added initial add_flow_enable implementation for V93K

ginty:
This update adds the ability for V93K flow modules to be generated with an enable-word wrapper, thereby allowing the top-level flow to easy enable/disable the execution of flow modules by a flow variable.

Currently, a top-level Flow.create call in Origen, will generate a V93K flow file with the following structure:

~~~text
group <FLOW_NAME> {
  // Flow content here
}
~~~

By enabling this feature in the target like this:

~~~ruby
OrigenTesters::V93K.new(add_flow_enable: :enabled)
~~~

the flow will now generate like this:

~~~text
group <FLOW_NAME> {
  if @<FLOW_NAME>_ENABLE == 1 {
    // Flow content here
  }
}
~~~

The @<FLOW_NAME>_ENABLE variable will be initialized to 1 in the generated variables file for the given flow, meaning that the module will run by default if the top-level flow does nothing with this variable.

Alternatively, the variable can be initialized to off like this:

~~~ruby
OrigenTesters::V93K.new(add_flow_enable: :disabled)
~~~

Meaning that by default, the module will not run and the top-level flow must always enable it by including @<FLOW_NAME>_ENABLE = 1 before hitting the module.

Setting this attribute in the target means that it will apply to all flows, alternatively it can be set within the flow itself (or the interface) if it is necessary to use this feature for only a subset of flows, or to override the default setting for a particular flow.
Here is an example:

~~~ruby
Flow.create interface: 'OrigenTesters::Test::Interface' do

  # Test that this can be overridden from the target at flow-level
  self.add_flow_enable = :enabled

  # Some functional test
  func :blah
end
~~~

This same API may be used to implement similar features on other platforms in future, but for now only the V93K is implemented.


redxeth:
@ginty We are definitely needing this I think as our 'customer' wants to have ability to have a single test flow for all test 'insertions'!  I assume defined via global variables like this.   Only thing I would add would be to permit specifying the name of the variable if you want-- at this time I don't know what naming convention my 'customer' wants but I can bet it will vary from your proposal.  Let me check.

ginty:
Going to go ahead and merge this, it is working well in production and we can always add the ability to customize the enable word name later if someone requires it.
Even if we add that ability, I think we would still want Origen to provide a sensible default name like this, so it seems safe to allow this to merge in its current state.

PR #42, opened 2017-06-12
Rfc2 - Updates to enable tester implemented overlay (and add DigSrc/Cap options)

pderouen:
Updates are nearly complete.  @ginty, can you add the 93k implementation?

The overlay API has changed to this:
~~~ruby
  tester.cycle overlay: {pins: dut.pin(:that_one), overlay_str: reg.bit[this_one].overlay_str}
~~~

The methods for getting "source_memory" and "capture_memory" might be able to be moved to a higher level than are currently (igxl tester/ base.rb).

Still need to kick the tires a bit to see if any further tweaks are needed.


pderouen:
Implemented the 93k stuff.  Appears to be working as expected.

ginty:
Hi @pderouen, sorry it took a while but I tried this out in one of my own apps and it looks good to me.

I confirmed that the V93K output is the same between old_testers + old_jtag, new_testers + old_jtag and new_testers + new_jtag.
I stepped through the JTAG code a bit and could see it using the new API, which all looks nice.

I took the liberty of making a few edits to move the common source/capture memory configuration API up to a higher level.

Apart from that, we just need some documentation (which doesn't go here anyway). If you are happy to sign off on the Teradyne side of things I think we can go ahead and merge this and release a new JTAG version.

pderouen:
@ginty, looks good to me.  Late last week I also added a transaction method to the api.  It simply yields and returns true.  This is avoiding needing to check tester.link? before using tester.transaction.

PR #44, opened 2017-07-31
Match2 keepalive updates

redxeth:
Adding capabilty to support 2-pin matchloop and keepalive to support unique test case on UltraFLEX.  This it bring UF up to what can be done on J750 already.

### DIFF
--- a/lib/origen_testers/igxl_based_tester/base.rb
+++ b/lib/origen_testers/igxl_based_tester/base.rb
@@ -331,7 +331,11 @@ module OrigenTesters
       #     $tester.end_subroutine
       def start_subroutine(name, options = {})
         local_subroutines << name.to_s.chomp unless local_subroutines.include?(name.to_s.chomp) || @inhibit_vectors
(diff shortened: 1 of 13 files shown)


redxeth:
@ginty let me know if this ok to merge into master then.  Would like to do tomorrow (Aug. 1).

lib/origen_testers/igxl_based_tester/base.rb
```diff
@@ -331,7 +331,11 @@ def call_subroutine(name, options = {})
       #     $tester.end_subroutine
       def start_subroutine(name, options = {})
         local_subroutines << name.to_s.chomp unless local_subroutines.include?(name.to_s.chomp) || @inhibit_vectors
-        microcode "global subr #{name}:"
+        if $tester.ultraflex? && name =~ /keep_?alive/
```
ginty: Should this also support option[:keep_alive]  ?

redxeth: Could add that.

PR #51, opened 2017-09-11
send timeset period to dut.current_timeset_period

pderouen:
At the moment origen_testers is not sending the current timeset or the current timeset period to the dut.  This PR is to send the timeset period to the dut to avoid run time errors when consuming the new timing api information via wave.evaluated_events.

@ginty I was hesitant to also send the timeset name to dut.current_timeset since this will cause a fail if timeset[id] hasn't been programmed.  How can this be done safely?

### DIFF
--- a/lib/origen_testers/timing.rb
+++ b/lib/origen_testers/timing.rb
@@ -146,14 +146,17 @@ module OrigenTesters
         original = @timeset
         timeset_changed(timeset)
         @timeset = timeset
+        # dut.current_timeset = timeset.name
         dut.current_timeset_period = period_in_ns
         yield
         timeset_changed(original)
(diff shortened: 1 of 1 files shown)


ginty:
Looks like I implemented it slightly differently on the stil_based_tester branch: https://github.com/Origen-SDK/origen_testers/blob/stil_based_tester/lib/origen_testers/timing.rb#L178

Not sure if that will merge any time soon so thanks for this PR.

Holding the name update off the way I did it and always updating the period seems like it should do the trick.

pderouen:
@ginty, thanks.  How's this?

ginty:
Included in 0.13.0

Issue #43, opened 2017-07-05
[V93K] Combine adjacent opposite if clauses


ginty:
The 93K flow generator should be smart enough to combine this:

      if @NVM_HOT_LIMITS == 1 then
      {
        run_and_branch(NVM_VMAX_VCMIN_BP_558EE47)
        then
        {
        }
        else
        {
          stop_bin "8778", "fail", , bad, noreprobe, red, 27, over_on;
        }
      }
      else
      {
      }
      if @NVM_HOT_LIMITS == 1 then
      {
      }
      else
      {
        run_and_branch(NVM_VMAX_VCMIN_BP_1_558EE47)
        then
        {
        }
        else
        {
          stop_bin "8778", "fail", , bad, noreprobe, red, 27, over_on;
        }
      }

into a single if statement like this:

      if @NVM_HOT_LIMITS == 1 then
      {
        run_and_branch(NVM_VMAX_VCMIN_BP_558EE47)
        then
        {
        }
        else
        {
          stop_bin "8778", "fail", , bad, noreprobe, red, 27, over_on;
        }
      }
      else
      {
        run_and_branch(NVM_VMAX_VCMIN_BP_1_558EE47)
        then
        {
        }
        else
        {
          stop_bin "8778", "fail", , bad, noreprobe, red, 27, over_on;
        }
      }

ginty:
Implemented in 0.13.0

PR #52, opened 2017-09-11
V93k flow optimizations

ginty:
This PR add the following V93K flow optimization improvements:

* Bug fix whereby the test name uniqueness signature was not appended to the first test in the flow.
* Improved the existing flag optimizer to make it recurse into nested groups and not only apply to the top-level
* Added a new adjacent if optimizer to address #43, you can see it collapsing an existing adjacent if condition here: https://github.com/Origen-SDK/origen_testers/compare/v93k_flow_optimizations?expand=1#diff-2f5b6d997b1e017e3d495f5553aff516L2675
* Added dedicated AST-level tests for the optimizers (plan to move these optimizers to the AST plugin at some point in the future)

Note: before opening this PR, the prb1 flow in master was updated to move the main flow content to an imported component, this exposed the fact that the existing flag optimizations where not applied to nested groups.
The majority of the diffs in the approved files are restoring these flag optimizations.


ginty:
Hi @rlaj, thanks for the initial review, could you give this a final look over please?

The github diff page isn't very helpful for reviewing the changes to the approved output.
Essentially all of the changes in the Teradyne output are just additional test cases which have been added to create additional optimization tests cases for V93K. I left in the the Teradyne build just to add additional tests, but it handles them fine.

The main diff in the V93K output is either new tests, or some cases where it now more efficiently optimizes than before.
Previously the flag_optimizer worked only on the top level, now it will recursively handle nested optimizations.

The main case of it is the two adjacent test cases starting here:
https://github.com/Origen-SDK/origen_testers/blob/master/program/flow_control.rb#L192

And here is the new output: https://github.com/Origen-SDK/origen_testers/blob/v93k_flow_optimizations/approved/v93k/flow_control.tf#L2506

It looks good to me and doesn't generate any flags, but would appreciate another pair of eyes on it.

Thanks!

arkohler:
Hi @ginty, I've taken the liberty to pull the branch and take a look at the optimizations on the flows I have.  On first glance this looks pretty good and I appreciate your hard work.  However, I am seeing a problem with some nested flows.  I'll put together a simplified generic test case tomorrow and post it here.

redxeth:
I am also seeing a problem with nested flows, import statements.   The third one is has stuff missing, trying to figure out why.  This is in my [fork ](https://github.com/redxeth/origen_testers) though I forked after @rlaj flow cleanup processor change.

Specifically when I import a subflow that has an import.

arkohler:
I'm seeing 3 issues with flow branching at this point.  This chunk of flow code would illustrate 2 of them
```
  log 'Nested optimization test case'
  func :outer_test, id: :ot
  if_failed :ot do
    unless_flag :flag1 do
      func :inner_test1, id: :it1
      render 'multi_bin;', if_failed: :it1
    end
  end
```
The output flow switches the compare order of "ot" and "flag1" which makes it so that the flow optimizer can't combine the "ot" fail flow into the test node.  The other issue is that the "if_failed: :it1" is not optimized at all for some reason.  **As far as I can tell, both of these issues still generate logically correct code.**
```
  print_dl("Nested optimization test case");
  run_and_branch(outer_test_BEA7F3B)
  then
  {
  }
  else
  {
    @OT_BEA7F3B_FAILED = 1;
  }
  if @FLAG1 == 1 then
  {
  }
  else
  {
    if @OT_BEA7F3B_FAILED == 1 then
    {
      run_and_branch(inner_test1_BEA7F3B)
      then
      {
      }
      else
      {
        @IT1_BEA7F3B_FAILED = 1;
      }
      if @IT1_BEA7F3B_FAILED == 1 then
      {
        multi_bin;
      }
      else
      {
      }
    }
    else
    {
    }
  }
```

The third issue is a bit more serious because it generates a flow that is not the same as the code.
```
  log 'Nested flag optimiztion test case'
  if_flag :flag1 do
    func :test4, id: :nf_t4
    if_failed :nf_t4 do
      render 'multi_bin;', if_flag: :flag1
    end
  end
```
This code completely optimizes away the second check for flag1.  Since we're checking the same flag for the same condition, I can understand why the optimizer is doing removing the second compare.  However, if flag1 is changed as a result of executing test4, the resulting flow will not behave as intended.  This also shows another example of a nested test not getting optimized.
```
  print_dl("Nested flag optimiztion test case");
  if @FLAG1 == 1 then
  {
    run_and_branch(test4_6_BEA7F3B)
    then
    {
    }
    else
    {
      @NF_T4_BEA7F3B_FAILED = 1;
    }
    if @NF_T4_BEA7F3B_FAILED == 1 then
    {
      multi_bin;
    }
    else
    {
    }
  }
  else
  {
  }

```

ginty:
Hi @arkohler, thanks for the additional test cases.

In all 3 cases the existing output was logically correct, but these were some additional opportunities for optimization which were not picked up, you can review the latest output for these test cases here:
https://github.com/Origen-SDK/origen_testers/blob/v93k_flow_optimizations/approved/v93k/flow_control.tf#L2873

On the 3rd point then, based on the available information in this code:

~~~ruby
log 'Nested flag optimiztion test case'
if_flag :flag1 do
  func :test4, id: :nf_t4
  if_failed :nf_t4 do
    render 'multi_bin;', if_flag: :flag1
  end
end
~~~

then I believe the current output is logically correct - i.e. you can't enter the outer do..end block without flag1 being set, so the compiler is correct to consider the inner `if_flag :flag1` condition redundant.

The optimization in question actually happens downstream in the ATP plugin output and it is not caused by these further V93K-specific optimizations being added by this PR, so I would propose we handle making changes to it separately.

However, we can have a think/discussion here about how we want to handle it. Essentially, for your use case the compiler needs to treat flag1 as volatile, so any references to it must be preserved and should not be optimized away.

I'm not sure if your use case is common enough that we should change the default behavior, but that is an open question and changing the current behavior is certainly one of the options on the table.
Here are the main options I can see:

- Consider any explicit references to a flag via `if/unless_flag` as potentially being volatile. Origen then should not perform any optimizations on these and only user flow variables (`if/unless_enable`) and Origen-generated flags would be optimized away where possible.
- As above, but only consider the flag volatile if it is not otherwise set by Origen APIs elsewhere in the current flow. i.e. if the flag looks to be owned by a 3rd party and not the current Origen flow.
- Provide an API for the application code to specify which flag(s) should be treated as volatile.

arkohler:
Thanks @ginty.  The output from the test cases looks much cleaner.
Regarding the volatile flags, I agree that it's not part of this v93k_flow_optimizations branch.  The master branch shows the same behavior.  I've submitted it as an issue to origen_testers #55.

ginty:
Included in 0.13.0

PR #54, opened 2017-09-15
Mixed case flags

arkohler:
I've added code to custom-process a "!" special character as the beginning of a flag name.
"FlagName" gets processed into "FLAGNAME" but
"!FlagName" is processed into "FlagName"

This requires the corresponding update to atp as well.


info-rchitect:
I think the use of '!' is ruby-ish in that it says 'take it as it is', similar to 'do it now'.  I know some sort of options hash is a typical method for this but given that the flow is a DSL, adding some intuitive syntax to the DSL works for me.

ginty:
I really appreciate these contributions from AMD, and this is a nice API, however I do have some reservation since `!FLAG` on Teradyne means when `FLAG` is not set.

I need to think about it a bit more, but that seems like it will be confusing. It is a primary design goal of Origen to have a program and pattern API that can re-target to any test platform.

info-rchitect:
yeah of course I miss the obvious of ! being negation!  And this is a convention I wholeheartedly support (the upcasing to prevent mistakes) but cannot find a workaround to currently.  Perhaps it could be an app specific option in the config/application.rb file?

~~~ruby
config.orgen_testers.flow_flags.upcase = false
~~~

arkohler:
I had considered a global flag implementation but I don't disagree with the initial intent of having uppercase flags.  I figured having a way to force literal strings would allow users to control the specific flags needed without affecting the auto-generated flags.  I can certainly understand the potential confusion of using "!", and we can easily change that to some other symbol.  My primary concern was trying to find some symbol that wouldn't appear in a valid flag name.  Perhaps "^", "%", or "#" would work better.  I'm not familiar with the syntax rules on the Teradyne testers so any input on that would be greatly appreciated.

ginty:
I think we should go for `$`.

One of the main advantages of that is that it is an acceptable first character in a symbol, so if you are used to using symbols for these then it will work too:

~~~ruby
func :test1, if_flag: :$Alarm

func :test1, if_flag: "$Alarm"
~~~

I thought for a second that maybe we should go down the path of a symbol being subject to the uppercasing while a string would be treated as a literal, but I think is already a large legacy code base which would not react well to such a change.

Before merging with that change I would also like to see a test case added for it, and ideally a doc line somewhere that mentions it.

Here are some pointers on how to create the test...

Run the tests by running `origen test`, or `origen examples` to run only the diff-based tests that we are going to use here.

There are test case programs in the program directory, add some code to one of them (flow_control.rb is probably a good choice), to exercise this feature. I think we want to cover the `if_flag` and `if_enable` flow control methods.

Make sure that the test is not restricted to only the V93K, we want to hit them all.

Run the tests again with this in place and review the log to see and review the diffs. If it looks good, copy the new version over to the approved directory.

In the case of the Teradyne output, you will initially see the `$` appear in the output, so you'll need to track down and modify the corresponding code in the Teradyne generator to make that go away.

Thanks!

arkohler:
Ginty, I think I've hit on all of your points with the this latest set of commits.
* Updated the PR to use a $ instead of a !.
* Added code into the flag processing routines of the IGXL output driver to remove any leading $
* Fixed a problem with set_flag: option in the original implementation
* Confirmed the same mixed-case flag functionality works with test enable flags
* Added a handful of tests to flow_control.rb and validated the output from all of the testers.
   - As part of this, several of the vars.tf files changed, but I'm not sure why.  The difference is only in the order of the variable definitions so functionality should be the same.

ginty:
Great stuff @arkohler, thanks for the updates.

I see the regression tests are failing: https://travis-ci.org/Origen-SDK/origen_testers/jobs/277075798

Are they passing when you run them locally?

arkohler:
Ginty, I believe the fail is due to the fact that it requires atp PR #4.  When I run both together locally the tests pass.

ginty:
Ok @arkohler, I've just release atp version 0.7.0 which includes your update.

Can you please bump the minimum atp version in origen_testers.gemspec and of that fixes everything we can get this released.

I've also added a note on this to the docs: http://origen-sdk.org/origen/guides/program/flowapi/#Manually_Setting_and_Referencing_Flags

arkohler:
I've updated origen_testers.gemspec to point to atp 0.7.0.  I let bundler pull the new version of atp, instead of using my local dev copy, and origen test passes.  Bundler also updated Gemfile.lock.

ginty:
Included in release 0.12.0.

@arkohler, enjoy the fruits of your labor, many thanks for the quality contribution!

arkohler:
Thanks @ginty, this was one small, but critical, point in merging origen-generated flows with our tester tools.

redxeth:
Sorry I'm late to the party y'all but one thing to NOTE-- the variables file template still upper cases all of the flags and so am definitely seeing a problem when I merged this master into my forked master.

I see a flag I'm using show up twice-- due to the fact that the template forces upcase-- but before the `clean_flow_control_variables` there appear to be 2 versions of one of my flags-- one uppercase and one lower case.   So this is a problem, will dig into it.

BTW please check [that fork](https://github.com/redxeth/origen_testers) if you like -- am still cleaning it up so not yet ready for a formal PR.  But just if you want to see what I'm up to.

arkohler:
@redxeth Now that you point it out, I'm seeing the same thing.  The issue is in the vars.tf.erb file.  I'll take a look.

redxeth:
Yeah --removed the upcase from teh template for debug-- would assume now that `generate_flag_name` would force all to uppercase since none of my flags have '$' but that doesn't appear to be the case. 
Isn't that what the default behavior is supposed to be?

arkohler:
@redxeth, in this PR, I was "fixing" the flag names as it printed out in the flow but the code builds two temporary arrays which are processed by the embedded ruby.  The fix in #56 preprocesses those flag names, with generate_flag_name, as they are placed into the arrays and removes the vars.tf.erb processing.  All of the test cases look clean.

PR #59, opened 2017-10-02
Integrate with ATP v1

ginty:
This is a significant update to integrate the changes in ATP v1 covered in this PR - https://github.com/Origen-SDK/atp/pull/6

It doesn't change the user facing API except to add some additional features, though it does make some changes to the output which are noted below.

Generally, this change simplifies the Teradyne and Advantest flow generators and pushes all the responsibility for AST optimization and transformation down to ATP.

This change provides the ability to mark flags as volatile to prevent them from being optimized away.

This test case:

~~~ruby
log 'Same test case with volatile flag'
volatile :$Alarm
if_flag :$Alarm do
  func :test4, id: :nf_t5
  if_failed :nf_t5 do
    render 'multi_bin;', if_flag: :$Alarm
  end
end
~~~

now generates this (without the volatile assignment the nested if $Alarm condition would be optimized away):

~~~
print_dl("Same test case with volatile flag");
if @Alarm == 1 then
{
   run_and_branch(test4_7_BEA7F3B)
   then
   {
   }
   else
   {
     if @Alarm == 1 then
     {
       multi_bin;
     }
     else
     {
     }
   }
}
else
{
}
~~~

It also adds a new API to define what should happen within branches via  blocks as shown in the examples below:

~~~ruby
log 'Test the block form of expressing if passed/failed dependents'
func :test1, on_pass: ->{
  func :test2
}, on_fail: ->{
  func :test3
  bin 10
}

log 'Test the else block on a flag condition'
if_enabled "bitmap", then: ->{
  test :test2
}, else: ->{
  test :test3
}
if_flag :some_flag, then: ->{
  test :test2
}, else: ->{
  test :test3
}
~~~

I have commented on all the flow diffs in the J750 and V93K output.

In the majority of cases these are not really by intention, but reflect the fact that the output is now more closely aligned to the way it was expressed in the source files - e.g. nested conditions now maintain the nesting order as expressed in the source file.

One notable exception to the above is an intended change to the way if_ran is implemented. It was found that the way it was implemented on J750/Uflex vs. V93K was different:

- On J750, a flag was set if the test passed, but the test was allowed to bin out unless it had been specifically set to continue on fail by the user
- On V93K, any bin out was removed and the flag was set by both passed and fail branches.

The J750 implementation was closer to the original intention of the if_ran condition, and its implementation has been changed to the following which follows the spirit of the J750 implementation while making it easier to implement across platforms:

- No changes are made to how the test bins out
- The RAN flag is now set by a dedicated operation as the very next thing after the test has executed
- For a group-level if_ran condition, the flag is now set immediately upon entering the group, rather than pushing it down to its contained tests

---

This is the original description for continuity with the original comments, but EVERYTHING BELOW THIS LINE IS SUPERSEDED BY THE ABOVE DESCRIPTION.

Integrates ATP version 0.8 which provides the ability to mark flags as volatile to prevent them from being optimized away.

This test case:

~~~ruby
log 'Same test case with volatile flag'
volatile :$Alarm
if_flag :$Alarm do
  func :test4, id: :nf_t5
  if_failed :nf_t5 do
    render 'multi_bin;', if_flag: :$Alarm
  end
end
~~~

now generates this:

~~~
print_dl("Same test case with volatile flag");
if @Alarm == 1 then
{
   run_and_branch(test4_7_BEA7F3B)
   then
   {
   }
   else
   {
     @NF_T5_BEA7F3B_FAILED = 1;
   }
}
else
{
}
if @NF_T5_BEA7F3B_FAILED == 1 then
{
   if @Alarm == 1 then
   {
     if @Alarm == 1 then
     {
       multi_bin;
     }
     else
     {
     }
   }
   else
   {
   }
}
else
{
}
~~~

Note the double `if @Alarm == 1` wrapping around the `multi-bin` call.

This is because of the way `if_flag :$Alarm do...end` is (currently) implemented at the lowest level. It is does not generate a branch in the AST, but instead registers the condition as currently open and then this gets attached to every test node that is generated within the block.

It would be quite a bit of work to change that, and hopefully the current implementation is good enough.

It does mean that the inline `if_flag: :$Alarm` is additional to the wrapping condition, so re-writing the source code to:

~~~ruby
log 'Same test case with volatile flag'
volatile :$Alarm
if_flag :$Alarm do
  func :test4, id: :nf_t5
  if_failed :nf_t5 do
    render 'multi_bin;'
  end
end
~~~

would probably give something close to the desired end result: 

~~~
print_dl("Same test case with volatile flag");
if @Alarm == 1 then
{
   run_and_branch(test4_7_BEA7F3B)
   then
   {
   }
   else
   {
     @NF_T5_BEA7F3B_FAILED = 1;
   }
}
else
{
}
if @NF_T5_BEA7F3B_FAILED == 1 then
{
   if @Alarm == 1 then
   {
     multi_bin;
   }
   else
   {
   }
}
else
{
}
~~~


arkohler:
Thanks for the work @ginty.  My concern is that this implementation eliminates some of the fail flag optimization, and in at least one case I can think of makes it difficult to generate logically correct code.

The reason we have volatile flags is that the execution of a test can change the value of the flag.  From your code snippet above:
```
volatile :$Alarm
if_flag :$Alarm do
  func :test4, id: :nf_t5
  if_failed :nf_t5 do
    render 'multi_bin;'
  end
end
```
If I knew nothing about the inner workings of Origen, I would expect the multi_bin to happen on a fail of test4, regardless of the state of @Alarm after executing test4.  By checking the flag again after test4, the output code is logically different than the expectation.  I'm not sure it's even possible write flow code doesn't recheck the volatile flag after test4.

The more I think about it, I believe @info-rchitect was onto something by suggesting that the volatile flags are really test flags rather than flow flags.  As an alternate implementation can I propose something like a test owner for the flag.  It seems to all behave as expected if the outer flag is different than the inner flag.  If the inner flag is treated a different flag, then the flow creator and optimizer might generate the right code with relatively little further modifications.
I've been thinking about the best way to do this.  You would need to add some data to the flag.  This could be something like appending ___<most_recent_test_id> and have anything with ___* stripped before printing.  You could also add some :owner attribute and volatile flags of different owners wouldn't get optimized.  I think you might even be able to get away with just keeping track of a nesting level rather than an owner.  Volatile flags of different nesting levels wouldn't be optimized, but the same level could be.
Thoughts?

ginty:
Yeah I see what you mean @arkohler.

Given that example source code, what should the output be in your opinion? Also, does the way it is being expressed at the source code level make sense, or is there a better way to write it?

If we can agree on those things first I think it will be easier to then talk about how to make it work.

arkohler:
For the original code:
```
log 'Same test case with volatile flag'
volatile :$Alarm
if_flag :$Alarm do
  func :test4, id: :nf_t5
  if_failed :nf_t5 do
    render 'multi_bin;', if_flag: :$Alarm
  end
end
```
I would expect it to look like
```
print_dl("Same test case with volatile flag");
if @Alarm == 1 then
{
   run_and_branch(test4_7_BEA7F3B)
   then
   {
   }
   else
   {
     if @Alarm == 1 then
     {
       multi_bin;
     }
     else
     {
     }
   }
}
else
{
}
```
In this case, I expect test4 to modify the value of @Alarm and I check it and bin accordingly.  I believe the original output is logically correct for this case, albeit somewhat convoluted.

Your second code snippet removes the inner @Alarm flag check so I would expect to see it removed from the flow accordingly:
```
log 'Same test case with volatile flag'
volatile :$Alarm
if_flag :$Alarm do
  func :test4, id: :nf_t5
  if_failed :nf_t5 do
    render 'multi_bin;'
  end
end
```
Should become
```
print_dl("Same test case with volatile flag");
if @Alarm == 1 then
{
   run_and_branch(test4_7_BEA7F3B)
   then
   {
   }
   else
   {
     multi_bin;
   }
}
else
{
}
```

To anticipate the question, "If test4 is altering the value of @Alarm, why wouldn't I want to check it?".  This becomes an issue because everything in the fail path inherits the outer flag check, but I have an instance where I want to run something in the fail path before the flag check.  We have the ability to set some variables in the software that persist across test methods and I need to temporarily set one of these variables to one value and reset it before binning.  I'm still checking the @Alarm flag, but not for every test in the if_failed flow.
```
log 'Same test case with volatile flag'
volatile :$Alarm
if_flag :$Alarm do
  set_var :test4_setvarA    # Sets some internal (non-flow) variable to value A
  func :test4, id: :nf_t5
  if_failed :nf_t5 do
    set_var :test4_setvarB    # Sets some internal (non-flow) variable to value B
    render 'multi_bin;' if_flag: :$Alarm
    unless_flag :$Alarm do
      func :test5
      render 'multi_bin;'
    end
  end
end
```
I would expect this code to come out logically like
```
print_dl("Same test case with volatile flag");
if @Alarm == 1 then
{
   run(test4_setvarA_BEA7F3B)
   run_and_branch(test4_7_BEA7F3B)
   then
   {
   }
   else
   {
     run(test4_setvarB_BEA7F3B)
     if @Alarm == 1 then
     {
       multi_bin;
     }
     else
     {
       run(test5_BEA7F3B)
       multi_bin;
     }
   }
}
else
{
}
```

ginty:
Thanks for the detailed reply @arkohler, that all makes sense.

I've decided to bite the bullet and overhaul the underlying representation of the flow in the ATP gem. The problem we are coming up against here is that a translation of the source code is done at the lowest level to turn it into something that is more like how a Teradyne flow is represented.
All conditions supplied in block form (do ... end) are not stored like that and rather are replicated to individually wrap each test contained within the block.
That makes it easy to generate a Teradyne flow where each test is represented as a single line which contains the test and all of the conditions that gate it.

However, by storing it like that at the lowest level, we have lost some fidelity and it is very difficult to reconstruct the condition branching back to exactly how the author expressed it. As you note, I think we can produce something functionally correct via this approach, but it really gets difficult to make it functionally correct and fully optimal.

With hindsight, we have gone down a wrong path by structuring the internal representation to ease Teradyne generation, and then go through a lot of additional processing to optimize it  for V93K.
The changes I am making will store it internally in the same structure as expressed in the source code.
That actually makes V93K generation much easier, and the structure we are trying to produce here will be trivial.
Then the transforms that we do to make it Teradyne friendly will be done higher up and only applied when targeting that platform.

While I am at it, I also plan to add support for the block syntax that you suggested here: https://stackoverflow.com/questions/46228502/origen-direct-flow-branching-control

Overall, this will be a very good change to make and will really improve the program generator foundations. I've been thinking that it would be good to do this for a while anyway, but I was hoping we might get away without it, however it is good that we have come up against this test case which is showing quite clearly that we need to address this weakness.

It will probably take a few days at least to get something working.

ginty:
Hi @arkohler, well this took a bit longer to fix than planned, but it was a worthwhile change to really overhaul the internal flow modelling with a less Teradyne-centric hat on.

I'll update this PR with more details and create a PR for the ATP updates tomorrow, but if you want to give this a try now you can.
Note that you will also need to pull the latest 'overhaul' branch of ATP into your application via your Gemfile for it to work.

Your test case now generates as you expect it: https://github.com/Origen-SDK/origen_testers/blob/volatile_flags/approved/v93k/flow_control.tf#L3323

approved/j750/flow_control_flow.txt
```diff
@@ -17,17 +17,21 @@ DFF 1.1	Flow Table
 						Test	pgm1	pgm1						Fail									flag-true	t4_C2A232F_PASSED				
 						Test	pgm1	pgm1						Fail									flag-true	t4_C2A232F_PASSED				
 						logprint	Test_that_if_ran_works																					
-						Test	pgm	pgm			10			Fail	t5_C2A232F_RAN													
+						Test	pgm	pgm			10			Fail														
+						flag-true	t5_C2A232F_RAN
```
ginty: Logically the same as the original, this reflects the new way of handling if_ran conditions by not changing the failing criteria for the target test and then setting a flag to indicate we got to this part of the flow immediately afterwards.

approved/j750/flow_control_flow.txt
```diff
@@ -114,24 +118,24 @@ DFF 1.1	Flow Table
 						Test	test4	test4						Fail									flag-true	ifallpb1_C2A232F_PASSED_AND_ifallpb2_C2A232F_PASSED				
 						logprint	Test_that_group-level_dependencies_work																					
 						Test	grp1_test1	grp1_test1			5			None		grp1_C2A232F_FAILED_0												
-						Test	grp1_test2	grp1_test2			5			None		grp1_C2A232F_FAILED_1												
-						Test	grp1_test3	grp1_test3			5			None		grp1_C2A232F_FAILED_2												
 						flag-true	grp1_C2A232F_FAILED																flag-true	grp1_C2A232F_FAILED_0				
+						Test	grp1_test2	grp1_test2			5			None		grp1_C2A232F_FAILED_1												
 						flag-true	grp1_C2A232F_FAILED																flag-true	grp1_C2A232F_FAILED_1				
+						Test	grp1_test3	grp1_test3			5			None		grp1_C2A232F_FAILED_2
```
ginty: Logically the same as the original, its just that setting the common flag based on the test-specific one now occurs next to each test rather than running all in a row at the end.

approved/j750/flow_control_flow.txt
```diff
@@ -114,24 +118,24 @@ DFF 1.1	Flow Table
...
 						Test	gt_grp1_test1	gt_grp1_test1			90			None		gt_grp1_C2A232F_FAILED_0												
-						Test	gt_grp1_test2	gt_grp1_test2			90			None		gt_grp1_C2A232F_FAILED_1												
 						flag-true	gt_grp1_C2A232F_FAILED																flag-true	gt_grp1_C2A232F_FAILED_0				
+						Test	gt_grp1_test2	gt_grp1_test2			90			None		gt_grp1_C2A232F_FAILED_1												
 						flag-true	gt_grp1_C2A232F_FAILED																flag-true	gt_grp1_C2A232F_FAILED_1				
 						Test	gt2	gt2			90			Fail									flag-true	gt_grp1_C2A232F_FAILED				
 						Test	gt_grp2_test1	gt_grp2_test1			90			None		gt_grp2_C2A232F_FAILED_0							flag-true	gt_grp1_C2A232F_FAILED				
-						Test	gt_grp2_test2	gt_grp2_test2			90			None		gt_grp2_C2A232F_FAILED_1							flag-true	gt_grp1_C2A232F_FAILED				
 						flag-true	gt_grp2_C2A232F_FAILED																flag-true	gt_grp2_C2A232F_FAILED_0				
+						Test	gt_grp2_test2	gt_grp2_test2			90			None		gt_grp2_C2A232F_FAILED_1							flag-true	gt_grp1_C2A232F_FAILED
```
ginty: See previous comment

approved/j750/flow_control_flow.txt
```diff
@@ -217,11 +221,11 @@ DFF 1.1	Flow Table
...
-						flag-true	at2_C2A232F_FAILED_AND_at1_C2A232F_FAILED																					
-						flag-false	at2_C2A232F_FAILED_AND_at1_C2A232F_FAILED															not	flag-true	at1_C2A232F_FAILED				
-						flag-false	at2_C2A232F_FAILED_AND_at1_C2A232F_FAILED															not	flag-true	at2_C2A232F_FAILED				
-						Test	test3	test3						Fail									flag-true	at2_C2A232F_FAILED_AND_at1_C2A232F_FAILED				
-						Test	test4	test4						Fail									flag-true	at2_C2A232F_FAILED_AND_at1_C2A232F_FAILED				
+						flag-true	at1_C2A232F_FAILED_AND_at2_C2A232F_FAILED																					
+						flag-false	at1_C2A232F_FAILED_AND_at2_C2A232F_FAILED															not	flag-true	at2_C2A232F_FAILED				
+						flag-false	at1_C2A232F_FAILED_AND_at2_C2A232F_FAILED															not	flag-true	at1_C2A232F_FAILED				
+						Test	test3	test3						Fail									flag-true	at1_C2A232F_FAILED_AND_at2_C2A232F_FAILED				
+						Test	test4	test4						Fail									flag-true	at1_C2A232F_FAILED_AND_at2_C2A232F_FAILED
```
ginty: Logically the same as the original, just the order of the names in the generated AND flag name is reversed, as is the order in which the flag is disabled based on one of the component flags being not set.

approved/j750/prb1_flow.txt
```diff
@@ -87,13 +87,17 @@ DFF 1.1	Flow Table
...
+						flag-true	erase_ran_2_7AF57D1_RAN																					
 						Test	margin_read1_all1_v1	margin_read1_all1						Fail									flag-true	erase_ran_1_7AF57D1_RAN				
 						Test	margin_read1_all1_v1	margin_read1_all1						Fail									flag-true	erase_ran_2_7AF57D1_RAN				
 						logprint	Test_unless_ran																					
-						Test	erase_all_v1	erase_all						Fail	erase_ran_3_7AF57D1_RAN													
-						Test	erase_all_v1	erase_all						Fail	erase_ran_4_7AF57D1_RAN													
+						Test	erase_all_v1	erase_all						Fail														
+						flag-true	erase_ran_3_7AF57D1_RAN																					
+						Test	erase_all_v1	erase_all						Fail														
+						flag-true	erase_ran_4_7AF57D1_RAN
```
ginty: More cases of the new if_ran changes

approved/j750/prb1_flow.txt
```diff
@@ -119,22 +123,26 @@ DFF 1.1	Flow Table
 						Test	meas_ps_leakage	meas_ps_leakage	1170		97		6	Fail														
 						logprint	Speed_binning_example_bug_from_video_5																					
 						Test	test200_1							None		g200_7AF57D1_FAILED_0												
-						Test	test200_2							None		g200_7AF57D1_FAILED_1												
-						Test	test200_3							None		g200_7AF57D1_FAILED_2												
 						flag-true	g200_7AF57D1_FAILED																flag-true	g200_7AF57D1_FAILED_0				
+						Test	test200_2							None		g200_7AF57D1_FAILED_1												
 						flag-true	g200_7AF57D1_FAILED																flag-true	g200_7AF57D1_FAILED_1				
+						Test	test200_3							None		g200_7AF57D1_FAILED_2
```
ginty: More cases of the common flag now being set closer to the responsible test.

approved/j750/prb1_flow.txt
```diff
@@ -119,22 +123,26 @@ DFF 1.1	Flow Table
...
-						Test	test100_1				5			Fail	g100_7AF57D1_RAN_0	g100_7AF57D1_RAN_0							flag-true	g200_7AF57D1_FAILED				
-						Test	test100_2				5			Fail	g100_7AF57D1_RAN_1	g100_7AF57D1_RAN_1							flag-true	g200_7AF57D1_FAILED				
-						Test	test100_3				5			Fail	g100_7AF57D1_RAN_2	g100_7AF57D1_RAN_2							flag-true	g200_7AF57D1_FAILED				
-						flag-true	g100_7AF57D1_RAN																flag-true	g100_7AF57D1_RAN_0				
-						flag-true	g100_7AF57D1_RAN																flag-true	g100_7AF57D1_RAN_1				
-						flag-true	g100_7AF57D1_RAN																flag-true	g100_7AF57D1_RAN_2				
+						flag-true	g100_7AF57D1_RAN																flag-true	g200_7AF57D1_FAILED				
+						Test	test100_1				5			Fail									flag-true	g200_7AF57D1_FAILED				
+						Test	test100_2				5			Fail									flag-true	g200_7AF57D1_FAILED				
+						Test	test100_3				5			Fail									flag-true	g200_7AF57D1_FAILED
```
ginty: This is a change to how if_ran is handled when it is applied to a group.
Previously the condition was applied to all contained tests, it is now far more simple and just sets the RAN flag immediately upon entering the group.

approved/j750/prb1_flow.txt
```diff
@@ -119,22 +123,26 @@ DFF 1.1	Flow Table
...
-						set-device					10			Fail									flag-true	alarm				
-						set-device					11			Fail								not	flag-true	alarm				
+						flag-true	sft1_7AF57D1_FAILED_AND_Alarm																					
+						flag-false	sft1_7AF57D1_FAILED_AND_Alarm															not	flag-true	Alarm				
+						flag-false	sft1_7AF57D1_FAILED_AND_Alarm															not	flag-true	sft1_7AF57D1_FAILED				
+						set-device					10			Fail									flag-true	sft1_7AF57D1_FAILED_AND_Alarm				
+						flag-true	sft1_7AF57D1_FAILED_AND_NOT_Alarm																					
+						flag-false	sft1_7AF57D1_FAILED_AND_NOT_Alarm																flag-true	Alarm				
+						flag-false	sft1_7AF57D1_FAILED_AND_NOT_Alarm															not	flag-true	sft1_7AF57D1_FAILED				
+						set-device					11			Fail								not	flag-true	sft1_7AF57D1_FAILED_AND_NOT_Alarm
```
ginty: This is an improvement, the original generation of this sequence was actually completely wrong and did not gate the set-device by the result of some_func_test.
The new output is correct.

approved/v93k/prb1.tf
```diff
@@ -1778,24 +1764,10 @@ test_flow
...
-      @ERASE_RAN_4_864CE8F_RAN = 1;
-    }
-    else
-    {
-      @ERASE_RAN_4_864CE8F_RAN = 1;
-    }
+    run(erase_all_33_864CE8F);
+    @ERASE_RAN_3_864CE8F_RAN = 1;
+    run(erase_all_34_864CE8F);
+    @ERASE_RAN_4_864CE8F_RAN = 1;
```
ginty: New implementation of if_ran, set the flag immediately after the test has run rather than affecting the test itself.

approved/v93k/prb1.tf
```diff
@@ -1838,11 +1810,11 @@ test_flow
...
-      }
-      else
-      {
-      }
-    }, open,"additional_erase", ""
+      }, open,"additional_erase", ""
+    }
+    else
+    {
+    }
```
ginty: Logically the same, just that the job context is now the outer one.
The new output aligns to the order the conditions are applied in the source file.

approved/v93k/prb1.tf
```diff
@@ -1887,35 +1859,30 @@ test_flow
     if @G200_864CE8F_FAILED == 1 then
     {
       {
+        @G100_864CE8F_RAN = 1;
```
ginty: New if_ran implementation for a group, set the flag immediately upon entering the group, rather than pushing it down to individual tests.

approved/v93k/prb2.tf
```diff
@@ -190,16 +190,16 @@ test_flow
...
-      }
-      else
-      {
-      }
-    }, open,"prb2_main_2", ""
+      }, open,"prb2_main_2", ""
+    }
+    else
+    {
+    }
```
ginty: Logically the same, just that the group is now inside the flag gate rather than the other way around.
The new output matches the way it is defined in the source file.

approved/v93k/flow_control.tf
```diff
@@ -3419,6 +3482,21 @@ test_flow
...
+  else
+  {
+  }
+  if @Alarm == 1 then
+  {
+  }
+  else
+  {
+    run(test2_27_BEA7F3B);
+  }
```
ginty: New test case that was added

approved/v93k/flow_control.tf
```diff
@@ -2413,12 +2455,12 @@ test_flow
...
   then
   {
-    @T8_BEA7F3B_RAN = 1;
   }
   else
   {
-    @T8_BEA7F3B_RAN = 1;
+    stop_bin "", "fail", , bad, noreprobe, red, 10, over_on;
   }
+  @T8_BEA7F3B_RAN = 1;
```
ginty: All cases of the new if_ran implementation where it doesn't affect the target test and now sets a flag after it has executed.

approved/v93k/flow_control.tf
```diff
@@ -3278,6 +3320,27 @@ test_flow
...
+        multi_bin;
+      }
+      else
+      {
+      }
+    }
+  }
+  else
+  {
+  }
```
ginty: New test case

approved/v93k/flow_control.tf
```diff
@@ -3325,9 +3388,9 @@ test_flow
     run(test3_15_BEA7F3B);
   }
   run(test4_9_BEA7F3B);
-  if @AF2_BEA7F3B_FAILED == 1 then
+  if @AF1_BEA7F3B_FAILED == 1 then
   {
-    if @AF1_BEA7F3B_FAILED == 1 then
+    if @AF2_BEA7F3B_FAILED == 1 then
```
ginty: Logically the same, just comes out with ordering matching the source file now.

approved/j750/flow_control_flow.txt
```diff
@@ -251,3 +255,15 @@ DFF 1.1	Flow Table
...
+						Test	test3	test3						Fail									flag-true	t114_C2A232F_FAILED				
+						set-device					10			Fail									flag-true	t114_C2A232F_FAILED				
+						Test	test2	test2						Fail									flag-true	t114_C2A232F_PASSED				
+						logprint	Test_the_else_block_on_a_flag_condition																					
+		bitmap				Test	test2							Fail														
+		bitmap				goto	LABEL_13_C2A232F																					
+						Test	test3							Fail														
+	LABEL_13_C2A232F					nop																						
+						Test	test2							Fail									flag-true	some_flag				
+						Test	test3							Fail								not	flag-true	some_flag
```
ginty: New test cases

approved/v93k/flow_control.tf
```diff
@@ -3419,6 +3524,49 @@ test_flow
...
+    run(test3);
+  }
+  if @SOME_FLAG == 1 then
+  {
+    run(test2);
+  }
+  else
+  {
+    run(test3);
+  }
```
ginty: New test cases

arkohler:
I found another test case that this PR fixes:

```
  log 'Test groups inside of a nested block'
  func :test1, id: :gt1
  if_failed :gt1 do
    group 'Test1 Fail flow' do
      func :test2
    end
  end
```

In the previous code the if_failed and group appeared to be processed in reverse order, breaking the test combiner.  In this PR it looks as I would expect it to.  I also confirmed it with the new lambda syntax and it looks correct.

Previous output:
```
  print_dl("Test groups inside of a nested block");
  run_and_branch(test1_29_BEA7F3B)
  then
  {
  }
  else
  {
    @GT1_BEA7F3B_FAILED = 1;
  }
  {
    if @GT1_BEA7F3B_FAILED == 1 then
    {
      run(test2_27_BEA7F3B);
    }
    else
    {
    }
  }, open,"Test1 Fail Flow", ""
```

Output from this PR:
```
  print_dl("Test groups inside of a nested block");
  run_and_branch(test1_31_BEA7F3B)
  then
  {
  }
  else
  {
    {
      run(test2_29_BEA7F3B);
    }, open,"Test1 Fail flow", ""
  }
```

This is similar to the "Test nested conditions on a group" test case, without passing conditions to the group itself.

ginty:
@Dieleman, could you please check this one out on your C402T app please?

@redxeth, @jiangliu23, @ThaoHuynhFsl, could one of you please check this out on your C90 app? Interested particularly in the Teradyne output and also to make sure it properly handles the power cycle injection.

To test it, you need to add the following to your Gemfile:

~~~ruby
gem 'atp', git: 'https://github.com/Origen-SDK/atp.git', branch: 'overhaul'
gem 'origen_testers', git: 'https://github.com/Origen-SDK/origen_testers.git', branch: 'volatile_flags'
~~~

When you are happy with it please approve this or else comment on any problems you find.

Thanks!

info-rchitect:
@ginty how much longer for a merge here?  We are waiting on this to update our own test interface, in preparation for our first internal demo of Origen.  Would like to close ASAP.  thx

ginty:
This is a definitely one of the bigger/riskier changes and I think we need the blessing from those I have tagged above.

I've asked them to comment here when they think they might be able to look at it.

redxeth:
@ginty I will make time to look at this week.

ThaoHuynhFsl:
Hi @ginty , it had been checked out on the J750 using our app and it works fine with no change in behavior of the flow.  It is good to go.

PR #61, opened 2017-10-02
V93k multiport

redxeth:
# Multiport Burst Capability

A **multiport burst** is the combination of one or more port bursts that are executed in parallel in a multiport test.   Multiport bursts are intended for being able to break down the pins of a DUT into port groups so you can ultimately implement concurrent testing.   

So a multiport burst (MPBurst) serves basically to call your Main burst.  For that reason we have to define these extra bursts in the PMFL and refer to them instead of the Main in the Test Suite.

To generate MPBurst for your tests-- simply define a MPort in your `$dut on_create`:

```ruby
def on_create
   ...
   $tester.multiport = 'portName'
end
```

To differentiate the new MPBurst, an extension to the pattern name is added.   By default this extension is 'pset', but can be changed by setting the `multiport_ext':

```ruby
def  on_create
   ...
   $tester.multiport_ext = 'mpburst'
end
```

The following output files are impacted:

- **Test Flow (TF)** -- seqlbl in Test Suite will refer to `<pattern>_<multiport_ext>` instead of `<pattern>`.
- **Pattern Master (PFML)** -- additional references to the MPBursts are added
- **Pattern Compiler AIV** -- multiport options used for Main patts and additinal MPBursts added.

## FUTURE WORK
Currently this only supports the case where you want to generate all your patterns using the same MP Port name.    Does not yet support if you want different MP Port name per pattern.


ginty:
Why is this adding so much additional approved output?
It seems like the change is pretty minor otherwise, would like to see this only generate a single flow or the minimum necessary to show the change.

Also there is no documentation, so at a minimum you should update this PR header to explain what new APIs you are adding here.
Thanks!

redxeth:
Each and every TF and PFML are changed-- as well as the pattern list file-- that's why have to have an additional dir.  Seems the best way so if person not using mutiport then then they can ignore  it.

ginty:
Why is generating one flow file, e.g. prb1.rb not sufficient to cover this change?

rlaj:
Should also support optional burst prefix, (e.g. 'mpb_').

redxeth:
I suppose it could be-- but I would prefer to have all of the modified stuff to be in its own folder-- it's how has been done for other updates been going on lately.  I don't see the big deal in having extra approved folders.

ginty:
The problem with having extra approved files is that any time there is a difference in them it makes PRs, like this one in its current form, very difficult to review.

You can still have your own folder, just don't think we need to be generating the entire test program to verify a change which should be visible in one flow.
Assuming you're referring to the v93 enable/disable folders, well those only contain the minimum set required, I needed two flows to verify each state and they don't contain the full program build.

However, I don't like them either and they will be coming back out again soon in favour of a spec-based approach as taken in this PR: https://github.com/Origen-SDK/origen_testers/pull/58/files

arkohler:
How much support does Origen have for v93k bursts?  This looks like it's just some AIV updates for the compiler and still calling a pattern name directly with the seqlbl field.

ginty:
Hi @arkohler, not much is the answer.

The only reason we are using multi-port is as a means of vector memory usage reduction - i.e. making patterns which only target the subset of required pins rather than them all, and this PR addresses that need.

More than happy for you to scope out and drive how we could add better support for MP, but this is not something that I see emerging out of our application space any time soon and you will need to lead on that if you want it.

PR #62, opened 2017-10-03
Polished the tf output to be a closer match to Smartest export

rlaj:
Added various sorting helper methods to mimic how Smartest exports the tf file on 'Save'.  Tested with several different test flows.  However, I'm certain there are still some corner cases not covered that will generate different spacing and possible ordering, but this PR should clean up most diff issues.

### DIFF
--- a/approved/v93k/custom_tests.tf
+++ b/approved/v93k/custom_tests.tf
@@ -1,81 +1,91 @@
 hp93000,testflow,0.1
 language_revision = 1;
- 
-information
- 
- 
-end
---------------------------------------------------
-implicit_declarations
+
+testmethodparameters
+
+tm_1:
+  "BadPractice" = "NO";
+  "Really.BadPractice" = "";
+  "myArg0" = "arg0_set";
+  "myArg1" = "a_default_value";
+  "myArg2" = "CURR";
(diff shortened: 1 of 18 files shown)


rlaj:
I assumed @ginty's changes would be transparent to these updates since they were in the lower-level ATP abstraction.  The tf rendering doesn't happen until later after all of the test flow logic has been built up and transformed into V93K specifics.  But @ginty may have other thoughts?

ginty:
No problem to merge this once it's passing from my perspective, agree that there is no conflict with my work.

approved/v93k/custom_tests.tf
```diff
@@ -1,81 +1,91 @@
...
-implicit_declarations
 
-end
------------------------------------------------------------------
 testmethodparameters
+
 tm_1:
-  "testerState" = "CONNECTED";
-  "testName" = "Functional";
+  "BadPractice" = "NO";
```
redxeth: What is this here?

rlaj: I think it's correct.  It's just re-ordering the parameters alphabetically, but the diff makes it look weird.

PR #65, opened 2017-10-05
changed label overlay behavior for J750 tester to local label

pderouen:
global label breaks backward compatibility of origen_jtag

### DIFF
--- a/approved/generic_overlay_capture/j750/test_overlay.atp
+++ b/approved/generic_overlay_capture/j750/test_overlay.atp
@@ -52,7 +52,7 @@ call dummy_str                                                   > tp0          
                                                                  > tp0                          1 1 H 1 001 ;
 call subr_test                                                   > tp0                          1 1 H 1 111 ; // overlay keeps
 repeat 20                                                        > tp0                          1 1 H 1 111 ;
-global label_test:                                                                              
+label_test:                                                                              
                                                                  > tp0                          1 1 H 1 101 ; // 1st line after global label for overlay
                                                                  > tp0                          1 1 H 1 101 ; // 2nd line after global label for overlay
                                                                  > tp0                          1 1 H 1 101 ; // 3rd line after global label for overlay
--- a/approved/generic_overlay_capture/j750/test_overlay.atp
+++ b/approved/generic_overlay_capture/j750/test_overlay.atp
@@ -1,23 +1,23 @@
 // ***************************************************************************
 // GENERATED:
-//   Time:    13-Jun-2017 09:59AM
-//   By:      pderouen
-//   Command: origen g pattern/tester_overlay.rb -t dut3.rb -e j750.rb
+//   Time:    12-Oct-2017 09:21AM
+//   By:      FSL\b04525
+//   Command: origen g .\pattern\tester_overlay.rb -t dut3.rb -e j750.rb
 // ***************************************************************************
 // ENVIRONMENT:
 //   Application
 //     Source:    git@github.com:Origen-SDK/origen_testers.git
-//     Version:   0.9.6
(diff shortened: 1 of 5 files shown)


pderouen:
@priyavadan  @redxeth  can you take a look at these changes to tester.overlay for label overlay?  Originally, all labels were global.  Now all labels are local unless global is requested.  Is this your expectation for J750 and uflex?

redxeth:
@pderouen why are we changing the default overlay behavior?  Can you give detail?

I think changing default behavior will break people's apps and should be avoided where possible.

pderouen:
@redxeth, tester implemented overlay is a new thing.  Previously origen_jtag handled overlay implementation.  It defaulted to local labels.  This change to local is to un-break apps and to un-break origen_jtag's regression test suite.

pderouen:
The corresponding origen_jtag PR, which is waiting for this PR:
https://github.com/Origen-SDK/origen_jtag/pull/5

pderouen:
In general I believe a local label unless global is requested is the correct behavior.  This is consistent with what was done originally and with the behavior of tester.label.

redxeth:
OK-- I see.  Not sure I agree with you there that local should be default-- it was my understanding to do overlays in VBT you had to use only global labels.   So I dont get why origen_jtag doesn't use them by default.  Seems the approveds in origen_jtag are the ones that need fixing.

pderouen:
Hmm, @rlaj ?

redxeth:
Ok per our offline convo-- then let me amend-- default label for **subroutines** should be global, which is mostly how I have used it.  The approved examples here are not subroutines, so I'm ok with the change (although the comments after should change too) assuming @rlaj is as well.

redxeth:
That said I would say that subroutine overlay examples that use global should be added to prove that they haven't been affected.

pderouen:
@redxeth I don't know how to get the link to the file and line number added as a hyper-link to my comment.  But, origen_testers, has a test for sub routine style overlay in pattern\tester_overlay.rb, lines 40-43.

The expected behavior of the testers plug-in is in approved/generic_overlay_capture.  You can see the LVM pattern being created will add an import statement for the subroutine and insert the call to it in the appropriate location.  Currently, the tester plug in doesn't generate the svm pattern... maybe at some point in the future it will.

redxeth:
@pderouen click on the line number then click on ellipsis that pops up.

image

pderouen:
@redxeth, thanks for the pointer.  I may have misunderstood the change request.  You want a test case using the global label option, is this correct?  I agree this should be added to the testers plug in.

redxeth:
Yes -- thought that global used by everyone for all overlays initially.  But if not- and global only really required by subr AND there isn't a test case then one needs to be added.  

Also put in a warning of some sort as you are changing the default behavior that users may have come to depend on.  At least until some future official revision occurs. (say something in the warning that the default will change so change your app accordingly).

pderouen:
@redxeth, I added tests for the global label style.

pderouen:
@redxeth, if we're happy with the implementation and coverage can this be approved and merged?

redxeth:
@pderouen, first silicon Monday and PC with IT now, but will try to take a look later today.

pderouen:
Well?  @priyavadan @redxeth can this be approved as is, any updates needed?  Should I try to find a different reviewer?

redxeth:
@pderouen on it... sorry been quite busy.  Doing a quick regression with a couple of patts with our main app.

pderouen:
Thanks @redxeth.  I'll wait a little longer to see if @priyavadan wants to weigh in and then merge if no changes are requested from him.

PR #72, opened 2017-11-15
V93K Limits File (RFC0004)

ginty:
This is PR is being opened early to capture discussion on this item as it is developed.

It implements the create limits file feature outlined in this RFC: https://github.com/Origen-SDK/rfcs/blob/master/text/0004-v93k-limits-file-support.md


ginty:
This initial version for comment does the following:

* Adds the tester.create_limits_file attribute, and when set...
* Creates one limits file per flow, in testtable/limits
* Removes all limits from the testflow file
* Sets all test binning in the flow to multi-bin

ginty:
@arkohler, @info-rchitect, the feedback we are getting from multiple product teams within NXP is that they all want the limits (and bin and test numbers) to be handled via a testtable limits file like we are implementing here.

Would I be right in thinking that you are also currently using such a limits file in your test program, but creating and managing it independently of Origen right now?

I'm just trying to gauge if there would be any objection to making this style of limits handling the default once it is available, and perhaps even phasing out support for the current in-line limits approach over time?

@rlaj, @chrisnappi, would appreciate your thoughts on this too.

ginty:
Continued over here due to the volatile_flags branch being merged - https://github.com/Origen-SDK/origen_testers/pull/73

PR #73, opened 2017-11-20
V93k limits file

ginty:
Implements the create limits file feature outlined in this RFC: https://github.com/Origen-SDK/rfcs/blob/master/text/0004-v93k-limits-file-support.md

This initial version for comment does the following:

- Adds the tester.create_limits_file attribute, and when set...
- Creates one limits file per flow, in testtable/limits
- Removes all limits from the testflow file
- Sets all test binning in the flow to multi-bin


ginty:
@arkohler, @info-rchitect, the feedback we are getting from multiple product teams within NXP is that they all want the limits (and bin and test numbers) to be handled via a testtable limits file like we are implementing here.

Would I be right in thinking that you are also currently using such a limits file in your test program, but creating and managing it independently of Origen right now?

I'm just trying to gauge if there would be any objection to making this style of limits handling the default once it is available, and perhaps even phasing out support for the current in-line limits approach over time?

@rlaj, @chrisnappi, would appreciate your thoughts on this too.

ginty:
As part of this change 3 possible ways for how to handle test failure are going to be supported:

1. Fail and bin out (normal behavior)
2. Don't care the failure, continue and don't bin (current behavior when `continue: true`) is supplied
3. Continue at the point of failure, but bin the device later / at the end of the flow (this is being added)

The last one is new and will be enabled by passing `delayed: true` to `flow.test`.

These 3 cases will be implemented as follows for V93K:

1. Overon: NO, multi-bin is called
2. Overon: ON, multi-bin is skipped
3. Overon: ON, multi-bin is called

The addition of `delayed: true` may be supported when targeting Teradyne platforms in the future, but for now it will just be ignored.

ginty:
As part of this change, some new options are going to be added to provide greater control over how failing tests are handled in the flow.

Generally, it seems like V93K users do not like tests to fail unless the device is going to be binned, this is because a failing test generates a 'red box' in the flow and this bubbles all the way up its parent groups, making potentially large sections of the flow look like they have failed when viewed from a high level.
Origen's normal way of implementing `:if_failed` type conditions will create such red boxes and users don't like this and want to see a "sea of green" for a passing device.
The existing way of handling allowed failures will be maintained, however a new way will be added which will register the test as passing but still set the `FAILED` flow flag. This is going to be opt-in rather than the default since it requires the application's test methods to support it (though the Origen stdlib will provide canned functions to pick up all the heavy lifting to implement it).

Additionally, with modern yield monitoring systems, there seems to be a preference for delaying binning, meaning that when a test fails the flow should continue to execute but remember and bin out the first failure at the end of the flow. So we need a way to express that tests should be handled like this.

Here is the proposed API and its effects on the V93K generation. Note that the Teradyne targets are not being updated at this time and they will simply ignore the new switches, however they may be updated to consider them in the future.

### :continue

To indicated that a test's failure should be ignored (not binned out), will continue to be by setting `continue: true`.
Specifying whether flags should be set on pass or on fail will still be done in exactly the same way as it is currently handled.

By default on V93K, this will cause the test to execute as normal but the bin (or multi-bin) call will be removed from the test's fail branch within the flow. Any on-pass or on-fail flags that require to be set will be done within the flow exactly as they are today.

A new attribute will be added to the base tester class to allow the user to indicate that they want all tests marked with `continue: true` to always pass:

~~~ruby
tester.force_pass_on_continue = true
~~~

When set to true, this will cause the limits for any test where `continue: true` to be forced to "NA", naturally making the test pass when its result is evaluated by the SMT APIs.
This allows monitor tests to be easily inserted into the flow via `continue: true` and you will still get the actual result in the STDF, but the box will always be green.

In the case where on-fail or on-pass flags are required to be set, the program generator will not insert them into the flow, but will instead supply them to the test method which must support attributes named `onPassFlag` and `onFailFlag`.
All test methods provided by the OrigenStdLib will support this, and the evaluate test result APIs provided by OrigenStdLib will automatically look for these and apply them when present. This will provide users with an easy way to integrate flow control handling into their custom test methods.

### :delayed

A new test option is going to be added:

~~~ruby
flow.test :my_test, delayed: true
~~~

On platforms which don't implement this, the test will bin out there are then.

On the 93K, this will cause the `overon` flag to be set for the test, which will mean that the flow will continue in the event of a fail and binning will be delayed until the end.

info-rchitect:
if you choose to support multiple limits I think Origen::Specs should as well.  I am working on a test implementation.

ginty:
@info-rchitect, yes we definitely want to do that.

What I was hoping for was to have a single object that encapsulated all possible values of a parameter spec, then we could pass the object to the program generator via `limits: <spec object>` and it would convert that into the multiple limits for the different test insertions.

I came to the conclusion that the current Spec API does not provide that, but I was hoping that this would come out in your overhaul of the Spec model/API.

info-rchitect:
@ginty the question we need to ask ourselves is whether ChipMode would suffice in describing the number of limits?  It does cover many of them, for example 'HTOL'.  But digging further, one may want a limits set based on lot type (corner vs nominal).  This means we would need to support the definition of arbitrary limits sets for a spec.  I think Origen should be opinionated in that the limits defined during instantiation should be the datasheet limits and any others added should be done via a new API.

~~~ruby
my_spec = spec :vol, :dc do |s|
  s.description 'output voltage'
  s.max = 0.2 * :vdd
  s.unit = 'V'
end

my_spec.limit_sets = [:datasheet]
my_spec.limits_sets[:datasheet].max.exp = "0.2 * :vdda"

my_spec.add_limit_set :corner_lot do |l|
  l.max = 0.5 * :vdda
end

my_spec.limits_sets = [:datasheet, :corner_lot]
my_spec.limits_sets[:corner_lot].max.exp = "0.5 * :vdda"
~~~

What I wonder about is whether the limits set 'context' should behave as  Origen::Paramters::Set.  At first I said yes but I am not really sure.  Where do Specs end and Parameters start from the viewpoint of programming something on an ATE or writing something in a document?

ginty:
@info-rchitect, getting a bit off topic for this PR, so would be good to continue this in a dedicated area, some comments though...

Ideally, I would like to see a low level spec model/API which represents spec variables using generic terms - selector, category and so on.
This could be based on the way STIL does it, or if we feel that is not rich enough we just expand it with additional selector attributes as we see fit.
I am not particularly concerned about aligning this to 3rd party standards, but it should be at least rich enough that it can contain all of the data that can be expressed in those standards. If the Origen model is richer though, that is fine, the main thing that we need here is a low level 'spec' model that everyone in the Origen eco-system can agree on and recognize.
I'm not sure if 'spec' is the right term for this lowest level of abstraction though, maybe `Origen::Model::Attribute` is better, meaning that it is an object used to represent some attribute of a model, which will encapsulate all the different values that that attribute can have when a different filter is applied to it. 'Selector' and 'Category' would be confusing at this level, what's the difference between them? So I would just go for something like `filter1`, `filter2`, etc. Probably we use Ruby's meta programming magic to put no hard limit on how many filter levels you can apply.

I think this level shouldn't even have the concept of a min and max value, rather a target and min/max for a given attribute should be represented by 3 `Model::Attribute` objects, `my_attr`, `my_attr_min` and `my_attr_max`.

In terms of representing a datasheet spec, then I see that as being an abstraction layer built on top of this. I think it should not be much more than an object which holds a collection of `Model::Attribute` objects to store the target/min/max, and a syntactic sugar API which makes the decisions that dc/ac will be captured by filter1, chip mode by filter2, etc.

I see the `Origen::Parameters` as being a low level element that is available to Origen application and plugin developers to use if it helps them, or not if it doesn't.

Should the `Model::Attribute` class use `Origen::Parameters` internally to make it work? Maybe if it helps, but if not I don't see any point in jumping through hoops to use it.

I don't see `Origen::Parameters` ever going away, it is a bounded piece of functionality which I would say is complete, but perhaps many application use cases of it will go away once we have a better Spec API available.
I think `Origen::Parameters` then would be something that app developers could use if they wanted something a bit more lightweight that the spec API, or if they fundamentally didn't agree with the way the Spec API does things, then it allows them to roll their own system within their application - which is how I have been using it so far.

Don't take any of this as gospel btw, I haven't fully thought it through and can't claim to fully understand the chip-level spec domain that you are trying to model. 

Conceptually though, having a generic low level which is the data store, then layers of abstraction built on top to tailor its use to a particular domain/application is the way we want to go I think.

info-rchitect:
@ginty thanks for the nice reply.  I will add it to the spec RFC coming soon.

ginty:
All reviewers - this has been sitting for a long time and works well in my application.
I need to build more stuff on top and the time has come to merge or object.

**Unless I hear otherwise, I am going to go ahead and merge this on the 2nd May - two days from now.**

info-rchitect:
@ginty fine by me.  WRt the Spec API, I have overhauled it and will make a PR soon.

coreyeng:
Hi @ginty, I don't really know enough about the V93K yet to have an opinion one way or the other. I think whatever you're doing is fine.

coreyeng:
LOL. I just realized I wasn't one the of the reviewers in the first place (definitely for the best). I just got the email and assumed I was. Nevermind me.

PR #84, opened 2018-03-20
Dynamic and looping testflow comment support

CodyBriscoe:
Allows for cc calls during test flow for dynamic comments. comments now use a source line lookup to transfer to the description to allow comments on looped tests


lib/origen_testers/api.rb
```diff
@@ -148,7 +148,18 @@ def c1(msg, options = {})
     def c2(msg, options = {})
       c1(msg, options)
     end
-
+    
+    def c3(msg, options = {}, line_number)
```
ginty: Per the comments here: https://github.com/Origen-SDK/origen/pull/250#issuecomment-375242047

I think this should be implemented as branch within the c1/cc method, rather than introducing a new method.

lib/origen_testers/flow.rb
```diff
@@ -47,7 +47,35 @@ def self.unique_ids=(val)
     def lines
       @lines
     end
+    
+    def self.ht_comments
```
ginty: Can you add the following comment to all of these methods please:

~~~ruby
# @api private
def self.ht_comments
~~~

That indicates that even although these are not private methods, they are not intended to be part of the user API and they will be marked as such in the generated API docs.

Here's an example from an existing method that I picked at random - http://origen-sdk.org/testers/api/OrigenTesters/VectorGenerator.html#remove_store_from_vector-instance_method

ginty:
Also you need to run 'origen lint' to clean up some style errors that are causing the build to fail.

Also, run 'origen examples' and 'origen specs' to check that they pass, it looks like this change is causing some failures.

Thanks.

CodyBriscoe:
Thanks for the feedback @ginty, I'll get on it

CodyBriscoe:
@ginty should be good to go now. I went ahead and updated all references for the examples as they looked good to me after updating _prb1_main.rb, but I can really only speak for the v93k files. Let me know if something else needs changing

PR #88, opened 2018-03-27
Patch 1

Dieleman:
Added boolean support and numeric formats don't get passed as string anymore


ginty:
Hi @Dieleman, thanks for the patch.

I think we need 'origen lint' run before we can merge this.

Also, I don't think this is compatible with SMT7, which could be handled by code on the application side, but perhaps better would be to allow the declaration as a boolean on the application side, then render true/false for SMT8 and 0/1 for SMT7.

What do you think?

PR #93, opened 2018-05-09
Fix case typo in ac_tml SpecSearch

rlaj:


lib/origen_testers/smartest_based_tester/base/test_methods/ac_tml.rb
```diff
@@ -18,16 +18,16 @@ class AcTml < BaseTml
...
+              method:         [:string, nil],
+              min:            [:string, nil],
+              output:         [:string, 'None', %w(None ReportUI ShowFailOnly)],
+              resolution:     [:string, ''],
+              result_pinlist: [:string, ''],
+              setup_pinlist:  [:string, ''],
+              spec:           [:string, nil],
+              step:           [:string, nil],
+              test_name:      [:string, 'SpecSearch_Test']
             }
```
ginty: Should we add aliases for these, may be a common mistake?

~~~ruby
  # ...
  test_name: [:string, 'SpecSearch_Test'],
  aliases: {
    result_pin_list: result_pinlist,
    setup_pin_list: setup_pinlist,
  }
}
~~~

PR #100, opened 2018-09-11
Add traditional pattern conversion proof of concept

ginty:
This PR adds the ability for Origen to function as a traditional pattern conversion tool.

It adds a global command that works like this outside of an Origen application workspace:

~~~text
origen convert path/to/my_pattern.stil -e v93k_smt7
~~~

When run outside of an app, an anonymous DUT instance will be created and its pins will be imported from the given pattern file.
The environment option defines the output, currently `j750`, `uflex` and `v93k_smt7` are supported.
`v93k_smt8` will also be added when that feature branch is merged (soon).

The command also works inside an application workspace, in that case the target and environment will work as normal to define the DUT and output format.

In theory then, it should also be able to feed into an OrigenSim/Link backend within an application, allowing an Origen app to provide a virtual test service to 3rd party pattern flows, although that has not been tested yet.

The only supported input format is currently STIL, with the level of STIL support dependent on the OrigenSTIL plugin. i.e. its probably not great yet, but please do raise tickets against OrigenSTIL with examples of anything you have tried and which it does not parse correctly.

I believe that @coreyeng is also working on importers for ATP and other ATE formats and these should hopefully be able to plug into this in the future too.

This change is actually fairly simple and is pretty much all contained in `lib/commands/convert.rb`.

The fact that It has been added is a reflection of the fact that the Origen ecosystem has now evolved to the point where adding this has been trivial.

It is up to the community now to see if they get value from this feature and to further drive its development.


info-rchitect:
@ginty very cool idea, have you done any benchmarking versus the 3rd party tool?  I know it will not be equivalent and that is not the purpose, just wondering.  Also, does Origen keep track of previously converted DUTs such that the DUT doesn't have to be recreated from scratch?

pderouen:
I could see a potential future use for a wgl converter.  Know of anything in the works?

coreyeng:
Yes, I'm working on a J750 and V93K ST7 importers (or at least the basics for them). Should be able to integrate those into this once complete. My decompiler API differs slightly but shouldn't be difficult to reconcile.

With the way the STIL gem was handled, do you think it worthwhile to split out the igxl/smartest testers into their own plugins and keep origen_testers as just a common API for implementing ATE-plugins to use?

@pderouen There was some internal discussion a year or two ago about WGL conversion but I don't think it ever went anyway. Stephen and I are trying to replace our internal tool so WGL conversion will need to occur somewhere, just don't think its been a priority yet.

ginty:
Any objections to merging this?

pderouen:
No objection from me.

coreyeng:
Looks fine to me too! FWIW, I'm still (slowly) working on #104 

Hopefully will have something to contribute to this before _too_ long.

PR #101, opened 2018-09-25
correct verbose inline vector comments with group_size > 1

pderouen:
When tester.vector_group_size is > 1 and the "-v" generate option is used, the inline cycle comment was incorrect. This fixes it. But, I couldn't figure out a way to explicitly test it. Any test suggestions?

### DIFF
--- a/lib/origen_testers/vector_pipeline.rb
+++ b/lib/origen_testers/vector_pipeline.rb
@@ -1,3 +1,5 @@
+# vector_pipeline concept explained [here](https://github.com/Origen-SDK/origen_testers/pull/101#issuecomment-424768720)
+
 module OrigenTesters
   class VectorPipeline
     attr_reader :group_size, :pipeline
@@ -130,11 +132,11 @@ module OrigenTesters
           yield vector
         end
         @vector_count += r
-        # @cycle_count += r unless @group_size > 1 && lead_group.last.repeat > 1
+        # @cycle_count now tracked in the calling methods (flush and empty)
       else
         yield vector
         @vector_count += 1
-        # @cycle_count += r * @group_size
+        # @cycle_count now tracked in the calling methods (flush and empty)
       end
     end
 
--- a/lib/origen_testers/vector_pipeline.rb
+++ b/lib/origen_testers/vector_pipeline.rb
@@ -124,7 +124,7 @@ module OrigenTesters
       else
         yield vector
         @vector_count += 1
-        @cycle_count += r
+        @cycle_count += r * @group_size
       end
     end
 
--- a/lib/origen_testers/vector_pipeline.rb
+++ b/lib/origen_testers/vector_pipeline.rb
@@ -89,19 +89,18 @@ module OrigenTesters
         end
 
         duplicate_last_vector until aligned?
-        ugly_i = 0
-        ugly_c = @group_size - 1
-        pipeline.each do |vector|
+
+        group_repeat_index = @group_size - 1
+        pipeline.each_index do |index|
+          vector = pipeline[index]
           vector.comments.each do |comment|
             yield comment
           end
           yield_vector(vector, &block)
-          @cycle_count += pipeline[ugly_c].repeat
-          ugly_i += 1
-          if ugly_i == @group_size
-            ugly_i = 0
-            ugly_c += @group_size
+          if index % @group_size == 0 && index > 0
+            group_repeat_index += @group_size
           end
+          @cycle_count += pipeline[group_repeat_index].repeat
         end
 
         comments.each do |comment|
--- a/lib/origen_testers/vector_pipeline.rb
+++ b/lib/origen_testers/vector_pipeline.rb
@@ -67,6 +67,7 @@ module OrigenTesters
             yield comment
           end
           yield_vector(vector, &block)
+          @cycle_count += pipeline[@group_size - 1].repeat
         end
         pipeline.shift(group_size)
       end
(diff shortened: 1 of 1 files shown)


lib/origen_testers/vector_pipeline.rb
```diff
@@ -124,7 +124,7 @@ def yield_vector(vector, &block)
       else
```
ginty: Is the same patch not required on line 123 also?

ginty:
No idea how to test this without adding additional logic to make it testable.

I think that's OK though since this is not a functional output and is essentially a cosmetic item.

pderouen:
Honestly, I don't fully comprehend what is going on in the group of yielding methods that yield_vector is a part of.  If I put in the group multiplier on line 123 the cycle count becomes too large. I did a little more testing and there's definitely a little more logic that is needed. I'm just not sure how to implement it.

What needs to happen is if the last vector in the group has a repeat, then line 123 needs to be skipped.  Otherwise, it should be +r.

ginty:
Hi @pderouen, here are some notes on the overall system here which might help with general understanding of this area either now or in the future (perhaps as part of this PR you could link to this in the comments at top of vector_pipeline.rb, or else copy it in as an in-line comment).

At pattern generation time, vectors are stored into a bank as they are generated - you can think of this as just an array of vectors.
As they are generated and pushed into the bank, no thought is given to optimization (combining identical adjacent vectors into a single vector with repeat) or implementation of dual or other multi-vector tester modes, they are simply pushed onto the end of the array as the application generates cycles.

At the end of the pattern, the (multi-vector mode aware) optimization is performed and the `vector_pipeline` picks up the heavy lifting for this.

The vector pipeline can be thought of as a black box where raw vectors are pushed in one end, and  optimized and mutli-vector aligned vectors come out the other.

The input to the pipeline is here - https://github.com/Origen-SDK/origen_testers/blob/6abc81bfd00346b576bae559abf810c22ca0be7e/lib/origen_testers/vector_generator.rb#L297

The output is via the `flush` method here - https://github.com/Origen-SDK/origen_testers/blob/6abc81bfd00346b576bae559abf810c22ca0be7e/lib/origen_testers/vector_generator.rb#L299

The `flush` method will yield if there are fully optimized vector(s) waiting to come out, otherwise it will do nothing.

It is beyond the scope of this comment to go through the pipeline implementation line-by-line, but here are some pointers to the key concepts.

The vectors at the head of the pipeline (the end closest to the output) are placed into what is known as the `lead_group`.
The number of vectors held by this group is the mutli-vector mode size - i.e. one by default, two in dual mode, four in quad mode, etc.

~~~text
# Pipeline in single mode:

* Empty       | Lead group
* Empty

# Pipeline in dual mode:

* Empty       | Lead group
* Empty       |
* Empty
* Empty
~~~

I'll use the dual mode for the rest of this example since that is what this PR relates to...

When the first vector is pushed into the pipeline it just takes up the first available slot, then same with the next vector. At this point a call to `flush` will do nothing since there may still be opportunity to incorporate the next vectors into these ones via a repeat:

~~~text
# pipeline << vec
* vec   R1    1011      | Lead group
* Empty                 |
* Empty
* Empty

# pipeline << vec
* vec   R1    1011      | Lead group
* vec   R1    1000      |
* Empty
* Empty
~~~

In the same way, the pipeline will swallow the next two vectors, still not yielding anything when `flush` is called:

~~~text
# pipeline << vec
# pipeline << vec
* vec   R1    1011      | Lead group
* vec   R1    1000      |
* vec   R1    1011
* vec   R1    1000
~~~

Now the caller here is regularly repeating the sequence, `pipeline << vec` followed by `pipeline.flush`.

Everytime `flush` is called it internally calls the method `lead_group_finalized?` which will kick off an optimization of the vectors currently in the pipeline.

In this case, we can see that the vector pairs are identical and therefore can be optimized into a repeat, so afterwards it will now look like this:

~~~text
# lead_group_finalized?     # => false
* vec   R2    1011      | Lead group
* vec   R2    1000      |
* Empty
* Empty
~~~

`lead_group_finalized?` has returned false in this case since we again back to our earlier state of only having one vector slot filled and with the prospect of being able to add further vectors into it.

So, we add some more:

~~~text
# pipeline << vec
# pipeline << vec
* vec   R2    1011      | Lead group
* vec   R2    1000      |
* vec   R1    1111
* vec   R1    1111
~~~

This time we cannot optimize further since the pin data has changed, so this time `lead_group_finalized?` will return true, `flush` will pop the lead group off the stack and yield it to the caller:

~~~text
# pipeline.flush
* vec   R1    1111      | Lead group
* vec   R1    1111      |
* Empty
* Empty
~~~

This process then repeats until no more vectors are left to be added to the pipeline, at this point a `pipeline.empty` method is called to yield any vectors that are still in there - https://github.com/Origen-SDK/origen_testers/blob/6abc81bfd00346b576bae559abf810c22ca0be7e/lib/origen_testers/vector_generator.rb#L306

The pipeline will also consider comments and microcode when considering whether vectors can be compressed together - fun fact, you can get smaller patterns therefore by turning off comments!

I haven't looked further into how to specifically solve your problem here, but hopefully this better understanding of the overall process will help you to arrive at a solution.

pderouen:
This works in more cases. I want to try a few more test cases.

pderouen:
@ginty can you take a look at this update? I think it will work correctly. Also, I included a link at the top of vector_pipeline.rb to the explanation of it. Is that the correct way to include the link? Haven't done that before.

pderouen:
@ginty, anything further needed on this PR?

lib/origen_testers/vector_pipeline.rb
pderouen: Had to move cycle tracking further upstream

pderouen:
~~~
               tp0   X  X  X;           |  First Vector of pair
repeat 5  tp0   X  X  X;           |  Second Vector of pair, a repeat here repeats the pair
~~~

yield_vector is only aware of the vector currently being processed. So, I changed flush and empty to grab the repeat count from the correct vector. I use that repeat count to increment the cycle count by the appropriate amount for all vectors in the group.

Yup, I'm happy with the changes. If we run across any more issues, I'll just open another PR.

ginty:
Sounds good, thanks.

PR #104, opened 2018-10-18
Pattern reader (Decompiler)

coreyeng:
Hello,

This is an initial PR for pattern decompilers (or reverse-compilers) that I've been hoarding in my workspace for a while. This is still a work in progress, but should give a decent starting point for either pattern conversion from third-party sources (e.g., scan) or for running a direct pattern source through simulation (scan, or patterns that are generated in one environment (windows) but run in another (Linux)). @ginty already has some third-party pattern source conversion in mind from #100 

This feature should be all standalone (was added around the existing code). I added some basic specs tests to show examples and help solidify the top-most API so other plugins can build on it without fear of it changing.

Currently, only ATPs and AVCs are supported. I also didn't differentiate between J750 and Ultraflex, but support for this could be added without too much trouble. I'm not too familiar with Ultraflex, so I'm not sure exactly of the differences and effects it will have. Smartest 8 was not considered. Only AVCs for Smartest 7 were tested. I'm not even sure if Advantest carried over support for AVCs to Smartest 8 or if it's something else now.

I tried to setup the grammars and nodes such that they could be used to support other platforms if needed. The grammars should give a pretty good starting point for future vector-based platforms, which can add or overwrite the base rules and base nodes.

Since this is pure Ruby, its decently slow. Even though I didn't go for optimization, the only ways to really speed this up will be to either write C extensions or decrease the number of rules. A warning on this: inspecting (printing the pattern model for) large patterns takes a very, very long time. My case for 'long pattern' was a 95k-line .atp. It parsed and symbolized in ~20 minutes (just on an average VM), but inspecting took an additional hour. Parse times actually seem to be more or less in line with our internal tools though I didn't account for memory usage. Using LSF provides additional speed boosts.

My target use case is to run pattern-direct simulation and, since this is just an initial version, there may be some missing features. For example, although I do parse microcode/firmware instructions, they aren't supported by OrigenSim (other than match loop), so I've ignored them during execution for now.

I didn't take into account 'origen' pattern styles. By this, I mean the pattern is taken at face-value, so the large

~~~
###########################
### COMMENT
###########################
~~~

Will decompile as three comments and is then itself 'commented' into either the simulation or converted pattern, leading to some messy comments.  Even though its messy, taking into account 'origen source' patterns I don't think is a needed feature at the moment. May be a good future item though and would help with OrigenSim comment cleanup at least.

I've been able to generate and run both J750 and V93K patterns and run them back through OrigenSim with passing results. That's what I'm using as my initial verification. Once the pattern converter is completed I'd like to add some examples and have the decompiler's regression based on taking approved patterns, 'converting' them to the same platform, then making sure they're functionally the same (ignore comments).

I'll also add some docs in the future, but I'd like to have the OrigenSim and pattern converter stuff done first so that there's more use cases available.


ginty:
Hi @coreyeng,

First off, thanks for yet another great PR!

I am really glad to see that you have taken the time to investigate and understand how these parsing grammars work as it was a bit of a concern to me that some of the plugins are starting to use them quite heavily but no one else apart from me understood how they worked.

While I appreciate the intent here is to commit an initial/baseline implementation, I think we should not rush to commit something that clearly needs improvement (i.e. > 1hr to convert is probably a non-starter for most folk). So we should view this initial PR state as the baseline, but iterate on it for a while here before merging it into master.

As you note, these parsers can be slow and are not best placed to handle large pattern files. The benefit they give you is the ability to create a robust parser that can handle the million-and-one-ways that a user could format and structure source code written in some rich programming language. However, for the most part, a tester pattern is a much more constrained format (e.g. there are not really many variations on how a vector can be formatted and they are guaranteed not to span more than one line) and so we are not getting that much benefit from going down the full grammar-parser route while still paying a heavy price for it in performance.

Therefore, I would recommend that you change this over to a hybrid approach whereby the pattern header section is parsed into an AST, but once it comes to processing the vectors it takes a more traditional line-by-line approach using a regex (which is how the STIL decompiler is implemented).

In fact, the STIL spec mandates that once you start defining vectors you are not allowed to do anything else, which basically is to make life easier on tools that are dealing with huge pattern files. i.e. you can parse the top section (what I called the _frontmatter_ in the STIL plugin) until you see the first vector, then you know that you can stop and have a full understanding of the structure and requirements of the pattern. Then, rather than having to read and parse the remainder of the pattern into memory at once, you can then drop to line-by-line processing/execution.

So I would like to see something similar here; for reference, here is the code from the STIL plugin which handles the vector-level processing - https://github.com/Origen-SDK/origen_stil/blob/master/lib/origen_stil/pattern.rb#L87

Once you have that, then the Origen conversion side of things becomes very easy (once you have setup the pins and so on based on what you parsed from the frontmatter) - you should use the `pin.vector_formatted_value = ` method to apply the values you read in from the vectors. See, https://github.com/Origen-SDK/origen_stil/blob/master/lib/origen_stil/pattern.rb#L25.

That is already a pretty optimized API and was created for a poor man's version of this tool from years ago which was taking ~10 minutes to convert a pattern by updating all pins via the conventional APIs but which dropped to seconds when using this method.

I would also like to see this follow the same API style as the STIL plugin, namely:

~~~ruby
OrigenTesters::J750.add_pins(file)   # Add the pins as defined in the given J750 .atp file

OrigenTesters::J750.execute(file)    # Execute the vectors from the given .atp
~~~

That would allow it to be easily used in the same way as the STIL plugin to include snippets of patterns from a 3rd party within Origen pattern source code, and also will make integration with the `origen convert` command really easy - https://github.com/Origen-SDK/origen_testers/blob/converter/lib/commands/convert.rb#L74

I do think it is absolutely fine to simply ignore any ATE micro code for now btw. The utility I really see for this is to allow 3rd party digital vectors, like mode entry sequences, to be able to be consumed and used in an Origen environment.

Of course once you have this basic structure setup, the vector parsing regex section can be enhanced over time to handle microcode references which have an Origen equivalent.

I think at this point though, is is important that we focus on getting the architecture right in terms of integration into the eco-system and performance/scalability.

Again, thanks for the significant effort you have already put into this and hopefully you don't find this feedback too disheartening.

ginty:
Hi @coreyeng,

First off, thanks for yet another great PR!

I am really glad to see that you have taken the time to investigate and understand how these parsing grammars work as it was a bit of a concern to me that some of the plugins are starting to use them quite heavily but no one else apart from me understood how they worked.

While I appreciate the intent here is to commit an initial/baseline implementation, I think we should not rush to commit something that clearly needs improvement (i.e. > 1hr to convert is probably a non-starter for most folk). So we should view this initial PR state as the baseline, but iterate on it for a while here before merging it into master.

As you note, these parsers can be slow and are not best placed to handle large pattern files. The benefit they give you is the ability to create a robust parser that can handle the million-and-one-ways that a user could format and structure source code written in some rich programming language. However, for the most part, a tester pattern is a much more constrained format (e.g. there are not really many variations on how a vector can be formatted and they are guaranteed not to span more than one line) and so we are not getting that much benefit from going down the full grammar-parser route while still paying a heavy price for it in performance.

Therefore, I would recommend that you change this over to a hybrid approach whereby the pattern header section is parsed into an AST, but once it comes to processing the vectors it takes a more traditional line-by-line approach using a regex (which is how the STIL decompiler is implemented).

In fact, the STIL spec mandates that once you start defining vectors you are not allowed to do anything else, which basically is to make life easier on tools that are dealing with huge pattern files. i.e. you can parse the top section (what I called the _frontmatter_ in the STIL plugin) until you see the first vector, then you know that you can stop and have a full understanding of the structure and requirements of the pattern. Then, rather than having to read and parse the remainder of the pattern into memory at once, you can then drop to line-by-line processing/execution.

So I would like to see something similar here; for reference, here is the code from the STIL plugin which handles the vector-level processing - https://github.com/Origen-SDK/origen_stil/blob/master/lib/origen_stil/pattern.rb#L87

Once you have that, then the Origen conversion side of things becomes very easy (once you have setup the pins and so on based on what you parsed from the frontmatter) - you should use the `pin.vector_formatted_value = ` method to apply the values you read in from the vectors. See, https://github.com/Origen-SDK/origen_stil/blob/master/lib/origen_stil/pattern.rb#L25.

That is already a pretty optimized API and was created for a poor man's version of this tool from years ago which was taking ~10 minutes to convert a pattern by updating all pins via the conventional APIs but which dropped to seconds when using this method.

I would also like to see this follow the same API style as the STIL plugin, namely:

~~~ruby
OrigenTesters::J750.add_pins(file)   # Add the pins as defined in the given J750 .atp file

OrigenTesters::J750.execute(file)    # Execute the vectors from the given .atp
~~~

That would allow it to be easily used in the same way as the STIL plugin to include snippets of patterns from a 3rd party within Origen pattern source code, and also will make integration with the `origen convert` command really easy - https://github.com/Origen-SDK/origen_testers/blob/converter/lib/commands/convert.rb#L74

I do think it is absolutely fine to simply ignore any ATE micro code for now btw. The utility I really see for this is to allow 3rd party digital vectors, like mode entry sequences, to be able to be consumed and used in an Origen environment.

Of course once you have this basic structure setup, the vector parsing regex section can be enhanced over time to handle microcode references which have an Origen equivalent.

I think at this point though, is is important that we focus on getting the architecture right in terms of integration into the eco-system and performance/scalability.

Again, thanks for the significant effort you have already put into this and hopefully you don't find this feedback too disheartening.

ginty:
Hi @coreyeng,

First off, thanks for yet another great PR!

I am really glad to see that you have taken the time to investigate and understand how these parsing grammars work as it was a bit of a concern to me that some of the plugins are starting to use them quite heavily but no one else apart from me understood how they worked.

While I appreciate the intent here is to commit an initial/baseline implementation, I think we should not rush to commit something that clearly needs improvement (i.e. > 1hr to convert is probably a non-starter for most folk). So we should view this initial PR state as the baseline, but iterate on it for a while here before merging it into master.

As you note, these parsers can be slow and are not best placed to handle large pattern files. The benefit they give you is the ability to create a robust parser that can handle the million-and-one-ways that a user could format and structure source code written in some rich programming language. However, for the most part, a tester pattern is a much more constrained format (e.g. there are not really many variations on how a vector can be formatted and they are guaranteed not to span more than one line) and so we are not getting that much benefit from going down the full grammar-parser route while still paying a heavy price for it in performance.

Therefore, I would recommend that you change this over to a hybrid approach whereby the pattern header section is parsed into an AST, but once it comes to processing the vectors it takes a more traditional line-by-line approach using a regex (which is how the STIL decompiler is implemented).

In fact, the STIL spec mandates that once you start defining vectors you are not allowed to do anything else, which basically is to make life easier on tools that are dealing with huge pattern files. i.e. you can parse the top section (what I called the _frontmatter_ in the STIL plugin) until you see the first vector, then you know that you can stop and have a full understanding of the structure and requirements of the pattern. Then, rather than having to read and parse the remainder of the pattern into memory at once, you can then drop to line-by-line processing/execution.

So I would like to see something similar here; for reference, here is the code from the STIL plugin which handles the vector-level processing - https://github.com/Origen-SDK/origen_stil/blob/master/lib/origen_stil/pattern.rb#L87

Once you have that, then the Origen conversion side of things becomes very easy (once you have setup the pins and so on based on what you parsed from the frontmatter) - you should use the `pin.vector_formatted_value = ` method to apply the values you read in from the vectors. See, https://github.com/Origen-SDK/origen_stil/blob/master/lib/origen_stil/pattern.rb#L25.

That is already a pretty optimized API and was created for a poor man's version of this tool from years ago which was taking ~10 minutes to convert a pattern by updating all pins via the conventional APIs but which dropped to seconds when using this method.

I would also like to see this follow the same API style as the STIL plugin, namely:

~~~ruby
OrigenTesters::J750.add_pins(file)   # Add the pins as defined in the given J750 .atp file

OrigenTesters::J750.execute(file)    # Execute the vectors from the given .atp
~~~

That would allow it to be easily used in the same way as the STIL plugin to include snippets of patterns from a 3rd party within Origen pattern source code, and also will make integration with the `origen convert` command really easy - https://github.com/Origen-SDK/origen_testers/blob/converter/lib/commands/convert.rb#L74

I do think it is absolutely fine to simply ignore any ATE micro code for now btw. The utility I really see for this is to allow 3rd party digital vectors, like mode entry sequences, to be able to be consumed and used in an Origen environment.

Of course once you have this basic structure setup, the vector parsing regex section can be enhanced over time to handle microcode references which have an Origen equivalent.

I think at this point though, is is important that we focus on getting the architecture right in terms of integration into the eco-system and performance/scalability.

Again, thanks for the significant effort you have already put into this and hopefully you don't find this feedback too disheartening.

ginty:
Hi @coreyeng,

First off, thanks for yet another great PR!

I am really glad to see that you have taken the time to investigate and understand how these parsing grammars work as it was a bit of a concern to me that some of the plugins are starting to use them quite heavily but no one else apart from me understood how they worked.

While I appreciate the intent here is to commit an initial/baseline implementation, I think we should not rush to commit something that clearly needs improvement (i.e. > 1hr to convert is probably a non-starter for most folk). So we should view this initial PR state as the baseline, but iterate on it for a while here before merging it into master.

As you note, these parsers can be slow and are not best placed to handle large pattern files. The benefit they give you is the ability to create a robust parser that can handle the million-and-one-ways that a user could format and structure source code written in some rich programming language. However, for the most part, a tester pattern is a much more constrained format (e.g. there are not really many variations on how a vector can be formatted and they are guaranteed not to span more than one line) and so we are not getting that much benefit from going down the full grammar-parser route while still paying a heavy price for it in performance.

Therefore, I would recommend that you change this over to a hybrid approach whereby the pattern header section is parsed into an AST, but once it comes to processing the vectors it takes a more traditional line-by-line approach using a regex (which is how the STIL decompiler is implemented).

In fact, the STIL spec mandates that once you start defining vectors you are not allowed to do anything else, which basically is to make life easier on tools that are dealing with huge pattern files. i.e. you can parse the top section (what I called the _frontmatter_ in the STIL plugin) until you see the first vector, then you know that you can stop and have a full understanding of the structure and requirements of the pattern. Then, rather than having to read and parse the remainder of the pattern into memory at once, you can then drop to line-by-line processing/execution.

So I would like to see something similar here; for reference, here is the code from the STIL plugin which handles the vector-level processing - https://github.com/Origen-SDK/origen_stil/blob/master/lib/origen_stil/pattern.rb#L87

Once you have that, then the Origen conversion side of things becomes very easy (once you have setup the pins and so on based on what you parsed from the frontmatter) - you should use the `pin.vector_formatted_value = ` method to apply the values you read in from the vectors. See, https://github.com/Origen-SDK/origen_stil/blob/master/lib/origen_stil/pattern.rb#L25.

That is already a pretty optimized API and was created for a poor man's version of this tool from years ago which was taking ~10 minutes to convert a pattern by updating all pins via the conventional APIs but which dropped to seconds when using this method.

I would also like to see this follow the same API style as the STIL plugin, namely:

~~~ruby
OrigenTesters::J750.add_pins(file)   # Add the pins as defined in the given J750 .atp file

OrigenTesters::J750.execute(file)    # Execute the vectors from the given .atp
~~~

That would allow it to be easily used in the same way as the STIL plugin to include snippets of patterns from a 3rd party within Origen pattern source code, and also will make integration with the `origen convert` command really easy - https://github.com/Origen-SDK/origen_testers/blob/converter/lib/commands/convert.rb#L74

I do think it is absolutely fine to simply ignore any ATE micro code for now btw. The utility I really see for this is to allow 3rd party digital vectors, like mode entry sequences, to be able to be consumed and used in an Origen environment.

Of course once you have this basic structure setup, the vector parsing regex section can be enhanced over time to handle microcode references which have an Origen equivalent.

I think at this point though, is is important that we focus on getting the architecture right in terms of integration into the eco-system and performance/scalability.

Again, thanks for the significant effort you have already put into this and hopefully you don't find this feedback too disheartening.

ginty:
Hi @coreyeng,

First off, thanks for yet another great PR!

I am really glad to see that you have taken the time to investigate and understand how these parsing grammars work as it was a bit of a concern to me that some of the plugins are starting to use them quite heavily but no one else apart from me understood how they worked.

While I appreciate the intent here is to commit an initial/baseline implementation, I think we should not rush to commit something that clearly needs improvement (i.e. > 1hr to convert is probably a non-starter for most folk). So we should view this initial PR state as the baseline, but iterate on it for a while here before merging it into master.

As you note, these parsers can be slow and are not best placed to handle large pattern files. The benefit they give you is the ability to create a robust parser that can handle the million-and-one-ways that a user could format and structure source code written in some rich programming language. However, for the most part, a tester pattern is a much more constrained format (e.g. there are not really many variations on how a vector can be formatted and they are guaranteed not to span more than one line) and so we are not getting that much benefit from going down the full grammar-parser route while still paying a heavy price for it in performance.

Therefore, I would recommend that you change this over to a hybrid approach whereby the pattern header section is parsed into an AST, but once it comes to processing the vectors it takes a more traditional line-by-line approach using a regex (which is how the STIL decompiler is implemented).

In fact, the STIL spec mandates that once you start defining vectors you are not allowed to do anything else, which basically is to make life easier on tools that are dealing with huge pattern files. i.e. you can parse the top section (what I called the _frontmatter_ in the STIL plugin) until you see the first vector, then you know that you can stop and have a full understanding of the structure and requirements of the pattern. Then, rather than having to read and parse the remainder of the pattern into memory at once, you can then drop to line-by-line processing/execution.

So I would like to see something similar here; for reference, here is the code from the STIL plugin which handles the vector-level processing - https://github.com/Origen-SDK/origen_stil/blob/master/lib/origen_stil/pattern.rb#L87

Once you have that, then the Origen conversion side of things becomes very easy (once you have setup the pins and so on based on what you parsed from the frontmatter) - you should use the `pin.vector_formatted_value = ` method to apply the values you read in from the vectors. See, https://github.com/Origen-SDK/origen_stil/blob/master/lib/origen_stil/pattern.rb#L25.

That is already a pretty optimized API and was created for a poor man's version of this tool from years ago which was taking ~10 minutes to convert a pattern by updating all pins via the conventional APIs but which dropped to seconds when using this method.

I would also like to see this follow the same API style as the STIL plugin, namely:

~~~ruby
OrigenTesters::J750.add_pins(file)   # Add the pins as defined in the given J750 .atp file

OrigenTesters::J750.execute(file)    # Execute the vectors from the given .atp
~~~

That would allow it to be easily used in the same way as the STIL plugin to include snippets of patterns from a 3rd party within Origen pattern source code, and also will make integration with the `origen convert` command really easy - https://github.com/Origen-SDK/origen_testers/blob/converter/lib/commands/convert.rb#L74

I do think it is absolutely fine to simply ignore any ATE micro code for now btw. The utility I really see for this is to allow 3rd party digital vectors, like mode entry sequences, to be able to be consumed and used in an Origen environment.

Of course once you have this basic structure setup, the vector parsing regex section can be enhanced over time to handle microcode references which have an Origen equivalent.

I think at this point though, is is important that we focus on getting the architecture right in terms of integration into the eco-system and performance/scalability.

Again, thanks for the significant effort you have already put into this and hopefully you don't find this feedback too disheartening.

coreyeng:
Thanks for the feedback @ginty!

I agree with most of what was said, but I'm a bit concerned with how long it may take to re-architect the parsers. I'm not familiar with AST, so it may end up being really easy to re-implement with it, or it might not be. Would we be able to meet in the middle for a merge onto master?

I'm not as concerned as the performance for a first release because:

* Inspecting (printing the full pattern model) is what takes significant time for larger patterns, not necessarily the parsing and symbolizing time. Obviously I only know our patterns, but my 'long' pattern was a 100k line ATP, which most patterns, even with ramload, tend to sit around 10k vectors. These can parse within about 5 minutes or less I think.
* From what I can see (though correct me if a I'm wrong), STIL isn't supplying an <code>inspect</code> anyway, so this feature wouldn't be missed it was just conveniently removed for performance issues for a bit.
* Although this does use a ton of RAM, the RAM usage and time-to-parse is comparable with our internal tool for pattern decompilation, and we continue to see improvement over this tool using OrigenSim, even with sub-optimal pattern parsing.
* (Assuming I fix the API) The details of how the decompilation works should be hidden from end users and as improvements are made, switching to AST from pure treetop shouldn't be an API change.
* There's no comparable feature for Origen (or in general, I think), so I'm not sure if we need to hold back a master for a feature that can be fixed later without affecting end users for just sub-optimal performance.

I'm more concerned about the API at this point, and I think I started looking at STIL but kind of forgot about it as I developed the ATP and AVC grammars, and I didn't consider pattern converter because it wasn't merged. Could we consider a merge onto master if I correct the user-API to conform to STIL's usage? I agree 100% with what you said regarding the API changes.

In general I wouldn't be too concerned about a master merge, but I have an app that's about to hit prime development time and another in pipeline that I'll need to start on and I'm concerned about trying to maintain branches during that time. I also want to add something similar to the pattern convert to OrigenSim, and need a OrigenTesters release in order to release an OrigenSim (something like origen sim pattern.atp, or something like that).

ginty:
@coreyeng, yeah if you fix up the API we can go ahead and re-work the internals for better performance later.

coreyeng:
@ginty, okay, sounds good. I'll start on that then and we can see how it looks. Upon just a quick inspection though, I have a few questions:

* How do you think <code>add_pins</code> and <code>add_pin_groups</code> would work in context of an ATP/AVC? I can add the pins/groups but I don't think I'd really have a choice between pin groups and pins. Put another way, if I have a <code>pta</code> pin group that's 12 pins long and a <code>reset</code> single pin in an ATP, I think I would have to add a <code>pta</code> pin group and a <code>reset</code> pin in order to get the pattern equivalent in Origen. How would this be handled? I could see just supplying a <code>add_pins</code> method and having it do both and then aliasing <code>add_pin_groups</code> to it?
* For <code>timesets</code>, since the AVC and ATP don't include timeset information I would need to look up the timeset from the DUT anyway. Do you think it worthwhile to just return the timeset instance instead of grabbing the <code>:time_in_ns</code> attribute?
* In <code>each_vector</code>, the <code>pattern_name</code> doesn't really make sense since ATPs and AVCs can only have one pattern per file. Would rather just leave that out and have the prototype be <code>each_vector(options = {})</code> or include a dummy parameter that will be ignored? For example: <code>each_vector(pattern_name_or_options = nil, options={})</code>
* Along the lines of vectors, the <code>Hash</code> return is fine, but we do have a nice <code>OrigenTesters::Vector</code> object that's built to handle a vector for multiple formats. Could we return these instead? Could also supply a <code>to_h</code> method for it if it doesn't already have one.

ginty:
Hi @coreyeng,

In general, the way that STIL is structured is that the top-level module is intended to define the public API - https://github.com/Origen-SDK/origen_stil/blob/master/lib/origen_stil.rb
The rest can be considered internal implementation detail.

- I agree, `add_pins` and `add_pin_groups` are probably just aliases in the context of an ATP/AVC and in fact per my above comment, the `STIL.add_pins` method does both anyway.
- Timesets is mainly internal detail. The `STIL.execute` has a `set_timesets:` options, which is false by default, which says whether it should apply the timesets as defined in the pattern. I think ATP/AVC should support this also (though I am fine with the initial implementation only supporting the default case of ignoring the timeset from the ATP/AVC).
- Each vector is internal detail, no need to align.
- I don't think you should be instantiating vectors at this level. Use whatever private data structures you want, but when it comes to generating vectors the API that you should ultimately use is `pin.vector_formatted_value=` and `tester.cycle`.

coreyeng:
@ginty, Ah! Okay, that makes sense. I was thinking the OrigenStil object had a public API as well. In that case, are really the only things missing just:

~~~ruby
OrigenTesters::J750.execute(pattern)
OrigenTesters::J750.add_pins(pattern)

OrigenTesters::UltraFLEX.execute(pattern)
OrigenTesters::UltraFLEX.add_pins(pattern)

OrigenTesters::V93K.execute(pattern)
OrigenTesters::V93K.add_pins(pattern)
~~~

I can fix up the timeset stuff as well. I think I just went with a base case that ended up working for me, but you're right, that needs to follow the ATP/AVC timesets. I'll fix that as well.

ginty:
Yep, I think that's all that is needed here

coreyeng:
Hi @ginty, if you get the chance, can you give this a quick look-through? I think I've matched the API closer to that of OrigenStil and I changed the specs to match the new API.

coreyeng:
Hi, are there any comments on the API updates? I'd like to get a release out to use in Origen Sim soon.

ginty:
Hi @coreyeng, yes I reviewed this in some detail yesterday and unfortunately I'm not going to approve it for merge at this time.
I know I said above that fixing the API should be sufficient, but after looking through it more closely I need to back track on that.

The main reason is that this is a complex piece of code which IMHO has a few fundamental issues, and I'm not convinced that they will get resolved later by turning a blind eye to them now.
There seems to me to be a high chance that this will remain with poor performance and/or it will be a shaky foundation on which more stuff is added.
I see this as being a premium feature that is being added to Origen's arsenal, I really don't want to say to users - "Come take a look, Origen can do pattern conversion now!", only for it to perform really badly and/or be very difficult to maintain.

At the end of the day, this is why we have peer reviews and its not acceptable to push it through just to make it available to a particular app. There are adequate ways to include non-official code in production apps if you need to, i.e. via a direct Git reference, and in fact many of my apps do this. Most notably the SMT8 branch of this same plugin which has not been merged to master yet exactly because I haven't completed the work to make it fully production-worthy.

So before merging this, these are the two key areas that we need to address:

- The parser implementation has no abstract syntax tree (AST) stage. A compiler is normally (and certainly this is true for all existing examples in Origen) comprised of three stages: parse/tokenization -> AST -> process/render. The compiler in this update missed out the AST stage and processes the tokenized representation directly. This makes it overall harder to do (this is why you are having to manually clean the tokenized representation), and it means that the process/render piece of it is done in a very ad-hoc way rather than being very formal/structured if done via walking an AST.

- The vectors need to not be formally parsed and update pin states via `pin.vector_formatted_value =` as already discussed earlier in this thread. I take your point that this is not an issue for the smaller patterns that you have in mind, but fundamentally we would be proceeding here with an architecture that has no chance of handling a scan pattern for example. Even with my proposed update it might still struggle and we might need to make it spool out directly to the pattern file rather than building everything up in memory. However, that would be an incremental update, whereas the only way to resolve this issue from here will involve significant re-architecture, and if we can't do that now, when will we ever be able to do it.

Having said all that, I don't expect that it needs to be sitting in limbo for the next six months, we should be able to work through the required updates within a few weeks.

I also don't think you need to do this in isolation, I am happy to start pushing updates to this branch to help - this is a feature that I really want to see implemented!

Again, I am sorry for misleading you on the way forward based on my initial review, and thanks again for the significant effort you have put into this so far.

coreyeng:
Thanks for the feedback @ginty! I agree with what was said, just not that they are a show stopper on a feature that doesn't yet exists. If you want to hold off until the performance is resolved, that's fine, I'm not just not sure when it'll be done. Though, as you've said, this will give more motivation to fix it now. I'll start on it and see where it goes.

Since it was previously said that fixing the API would be sufficient and since I felt I had done that with these latest commits, I was going to push it through on those merits. But if that's not true anymore then I won't merge obviously. I've been using unofficial releases for this for some already anyway.

I think I missed the section on <code>pin.vector_formatted_value =</code> though. I took that directly from OrigenStim. I could see skipping over that for a direct pattern to pattern conversion but I don't know of another API available to simulate a pattern without that. I read the other posts as saying to use <code>pin.vector_formatted_value =</code> so I'm a bit confused now.

ginty:
On the comment re. the feature not existing yet, just to say that by merging to master we are effectively agreeing to support said feature for ever more, so one eye always has to be on the future maintainability of what we are allowing in, and for me this update has raised a red flag in that area.
I wouldn't say that the issue is on the performance per se, if there was an incremental path we could take to get better performance later on then I wouldn't have an issue with it, but here we are talking about significant ground up changes to get there and that will be very hard to do in future (given the large surface area of this feature) unless we address it now.

Not sure exactly how to resolve the confusion around `pin.vector_formatted_value = `, let me know if this doesn't do it...
The main benefit of using this API is that it skips making the translation of `H` (for example) to `read + 1` and returns immediately if the pin is already at the given formatted value. If you call `pin.read(1)` then it will internally go through the translation of that back to `H` and also go on to do additional things like blow affected pin group value caches.
If you combine that with a local handle on the pin objects themselves (so that they don't need to be retrieved from the store by name on each vector) then it is a very performant way of translating vectors - especially if the source pattern contains pins groups (since `pin_group.vector_formatted_value=` is a very quick way to process multiple pins).
It does resolve to a call to a tester API to update the pin value from a pre-formatted state, so potentially OrigenSim may need a minor update to implement that, but from my perspective using the `vector_formatted_value=` API at this level should be compatible with all back end 'testers' - https://github.com/Origen-SDK/origen/blob/master/lib/origen/pins/pin.rb#L304

So how do we want to proceed here?

One route would be for me to push some updates to this branch to move it in the right direction, that would probably break the tests (which are great btw), but from that point you could then step in to work us back towards a passing state.

coreyeng:
Hi @ginty, sorry for the long silence. I'm working on this in the background, but no real ETA. I'm on branch of this branch, so we can continue to use this one for a 'first attempt', in case changes are needed to the grammars or the decompiler, since I do have some live applications using this.

I'll close this PR and open a new one when its ready. Thanks!

PR #107, opened 2019-01-18
Misc testflow formatting (V93K)

rlaj:
* Testflow indention spacing to match ST-saved .tf files
* Removal of testmethodlimits, testmethods, or test_suites sections if they are empty (follows ST-saved format)


info-rchitect:
@arkohler Can you review this please?  I am not an expert on the V93K formatting but from here it looks good.

ginty:
What's ST-saved ?

rlaj:
If I import the Origen-generated testflow file into SmartTest, then make a minor change and Save (from the GUI).

ginty:
Ah, yes seems sensible to align to that!

rlaj:
Good catch, @arkohler .  Patched it.

PR #112, opened 2019-01-30
allows users to not reload target when importing sub flows

info-rchitect:
not much to say, allows us to shorten our test flow generation time.


info-rchitect:
@ginty How do I add spec tests for this?

ginty:
Not easy to test, though don't like this implementation anyway per comments on the other piece of it.

If the downstream bit is changed to reverse the logic, this part should not be required.

info-rchitect:
going to start over

PR #114, opened 2019-01-30
needed to make site_control and site_match programmable

info-rchitect:
There are some search test methods that cannot be run in parallel mode.


info-rchitect:
@ginty How do I add spec tests for this?

ginty:
Nice update, I guess the default case is already covered by the existing tests.

To create a test for a non-default value you could update any of the tests in the flows in `program/` to pass in an option, then update `lib/origen_testers/test/interface.rb` to pick up that option and set the alternative value.

Then run `origen examples`, review the change and save the new approved file.

PR #117, opened 2019-02-11
print svm_subr optional for non uflex only.

williamcheong:
the microcode printout "svm_subr" only for IGXL of J750 and Flex, but not UltraFlex.
Thus bypass this print works well too, since in Uflex pattern compiler there's an "import undefineds" option to overcome this.


pderouen:
Hi William, I think the default behavior for Origen should be to add the subroutine import line. The goal is to make things as easy as possible for the user. Is there a use case where you have a call to a subroutine and having the import line is incorrect?

williamcheong:
@pderouen i agree with you. Had looked into the opcode, actually recognized by pattern compiler and passed IGXL tp validation. Just keep as it is should be alright. Thanks.

PR #120, opened 2019-05-16
Adds support for concurrent pattern generation

ginty:
Adds some minor hooks required to support the concurrent pattern generation feature that has recently been added to Origen core and which is documented here - https://origen-sdk.org/origen/guides/pattern/concurrent/

These updates do the following:

- Injects a thread ID into all pattern comments when running multi-threaded/concurrent mode.
- If `tester.cycle` is called within a child (i.e. not the main) thread, then it hands over to a `cycle` method implemented by the pattern sequencer, otherwise if called from the main thread in either single or multi-threaded mode it will be processed as normal.
- Bumps the min Origen version to `0.44.0` which supports the necessary `PatSeq` API.


pderouen:
@ginty these changes look fine to me.

PR #121, opened 2019-05-16
add uflex support for multiple use-limit lines with programable tname

pderouen:
Pretty simple addition of some functionality. A test instance can make multiple measurements and test all of them using a "Use-Limit" entry on the flow sheet. But, there was previously no way to specify individual test names for each use-limit entry.


pderouen:
@rlaj let me know if the functionality created here is the equivalent of your multi-limit branch.

priyavadan:
@pderouen great job on the PR, looks good to me. Please add some documentation for this feature in the origen_guides section (part of origen_testers/templates)

ginty:
Hi @pderouen, thanks for the PR and great to see you moving into this area of Origen!

I've approved your updates, but I wanted to point out to you some of the general direction in this area since it is not obvious and poorly documented at present...

The limit API you are enhancing here is not officially deprecated, but I do consider it a legacy way of doing limits, with the big weakness being that it is platform specific.
That means that it will either be ignored by downstream tooling (e.g. documentation generators, offline runners/simulators, and of course re-targeting to another platform), or else it will drive similar platform-specific handling onto them too.

The program generator evolved in this order: J750 -> UFlex -> V93K, and when adding support for the UFlex it seemed like it was OK to treat multiple limits as a quirk of that platform and add a custom API that is implemented within the UFlex generator classes.
Then the 93K came along which also supported multiple limits and with a completely different syntax program architecture which drove a re-think of the whole generator.

At this point it became clear that program generation is actually a *much* harder job than pattern generation. It really is like the wild west with vendors all doing completely their own thing, whereas on the pattern side you at least have the foundation that a vector is a vector and a cycle is a cycle regardless of what they called and how they are formatted.

So Origen could have continued down the path established by the early Teradyne generators and similar to how patterns are handled, whereby there is a common API supplemented by platform specific APIs, or it could move towards a more formal compiler architecture. Ultimately the latter was chosen and where we are currently is somewhere in between as features transition over to the formal compiler over time.

This means that in the absence of a formal standard of how test programs are described, Origen is creating one and this is handled by the Abstract Test Program (ATP) gem - https://github.com/Origen-SDK/atp
This is an example of AST (abstract syntax tree) from compiler theory and the flow we are moving towards is: user APIs generate the AST, and then platform-specific processors convert the AST into the equivalent IG-XL or SmarTest representation. Hopefully you can imagine that having that common internal representation means that the frontend user API can really be commonized and the  job of rendering the program as a platform-specific implementation, or as a common view like  documentation is made much easier.
We do plan to move ATP into OrigenTesters btw since having them separate is proving to be a hassle as program generator updates often end up being comprised of two related PRs.

Most of the Uflex program generator does not use the AST, however the flow does - https://github.com/Origen-SDK/origen_testers/blob/master/lib/origen_testers/igxl_based_tester/base/flow.rb
You will see that it is a sub-class of `ATP::Formatter` and what that provides is the infrastructure to systematically process the AST (so-called walking the AST), and allow the given formatter to define how each node type should be formatted/handled. So all methods that you see in there with the format `on_<type>(node)` will be called whenever the given node type is encountered and the formatter decides what to do with it. For example, whenever `on_test(node)` is called the IG-XL formatter/generator will create a new flow line and populate it with the information contained in the node. You can read some more about that here (this is the gem we use under the hood) - http://whitequark.github.io/ast/AST/Processor/Mixin.html

It is important to note that there is only one AST and it should semantically describe the complete test program, i.e. there is no 'flow AST', 'test instances AST' and so on.
In the case of something like limits, the AST represents them within Test nodes since that's what they conceptually apply to. However the flow generator for IG-XL would just ignore them, but the test instance generator (when walking the same AST) would pull them out. SmarTest likes to have all the limits in a dedicated limits file, so again we can do that easily now by having that particular generator ignore almost everything in the AST except for the limits.

ATP does already support representing the kind of thing you are wanting to do here, a limits option when creating a test can be added like this:

~~~ruby
test :test1, limits: [{ value: 1.25, rule: :lt}]
test :test2, limits: [ { value: 1, rule: :gt, units: :mV }, { value: 5, rule: :lte}]
~~~

and sub-tests (with their own limits) can be described like this:

~~~ruby
test :test1, 
  sub_tests: [
    sub_test(:test1_s1, limits: [{ value: 5, rule: :lte}, { value: 1, rule: :gt, units: :mV }]),
    sub_test(:test1_s2, bin: 10),
]
~~~

You can see how this ends up getting represented in some of the ATP spec tests - https://github.com/Origen-SDK/atp/blob/d225c1f29b345ca3a1485257adb84a24441c3f1b/spec/test_node_spec.rb#L112

Anyway, I think it is not fair to put the responsibility for conversion to this approach onto this PR!

However, I really hope that you will become active in the future development of the program generator, so I wanted to make you aware of the bigger picture here and hopefully you can bear that in mind when thinking about future updates.

Again, thanks for this PR and great to see you here!

pderouen:
@ginty yes I consider this just a patch to the existing uflex flow generation. It would be far better to have the same abstract description used across platforms. I'm glad something exists. It makes for a better experience.

@priyavadan thank you for pointing me to the documentation source. I'll try to add something today. As I mentioned it won't be comprehensive. But, better than what's there now.

pderouen:
@priyavadan, @ginty the uflex flow builder was duplicating the test number when multiple limits are applied. This would violate the stdf standard. I updated the code to increment the test number. This is a change in behavior though. 

Would it be better to allow the test numbers to be passed in an array and retain the original behavior if no limit test numbers are supplied?

ginty:
Rather than adding yet more stuff to the legacy API, I think it is worth considering adding support for the newer one.

This spec test:

~~~ruby
functional :test2, sbin: 101, number: 20020, lo: [-2, -1], hi: [2, 1], limit_tname: ['lim1', 'lim2']
~~~

would be written as:

~~~ruby
functional :test2, sbin: 101, number: 20020, 
  sub_tests: [
    sub_test(:lim1, lo: -2, hi: 2, number: 20021),
    sub_test(:lim2, lo: -1, hi: 1, number: 20022),
  ]
~~~

I'm not sure if the sub_test helper accepts limits in the lo/hi format instead of the more formal declaration from my earlier comment, but it should be easy enough to add in support for that shorthand syntax.

Since this seems to be handled in the flow sheet anyway for IG-XL, then its already structured to work with the AST, and you can just add an `on_sub_test` handler and have it render the required flow line.

pderouen:
@ginty sure that's fine. I don't have any experience with this object type though. How do I get at the data in it and what data is there supposed to be?

~~~
(byebug) node
s(:sub_test,
  s(:object, "lim1"),
  s(:limit, -2, "gte", nil, nil),
  s(:limit, 2, "lte", nil, nil))
~~~

I can hack up something really ugly, but surely there's a better way than this:

~~~ruby
  tname = node.find(:object).value
  limits = node.find_all(:limit)
  # grab each value and figure out which is low/high, don't know how to get at the "gte"/"lte"
~~~

ginty:
@pderouen, ok great and no problem with providing consultation on it...

So when implementing a handler for something like this `sub_test` node, you have the choice whether to manually extract and handle sub-nodes, like the limits here, or whether to only handle immediate properties and then let processing continue for other handlers, like `limit` node handlers to deal with the sub-nodes.
I think there is no hard and fast rule over which approach to take, and this is something that you will get a feel for the more you work with AST processors.

I'll describe both approaches here in a minute.

With regards to the node structure, you can assume that where a node contains endpoint values (like 2, "gte" in this example)  rather than a collection of nodes, then the order is fixed. Usually the order would be fixed even if it contained only other nodes, but sometimes in that case you might find that the presence of some node types may not always be guaranteed.

The structure of the ATP::AST should be defined in documentation, but alas it is currently lacking so you will need to dig into the source code. Most if not all nodes are currently generated with the `ATP::Flow` class - https://github.com/Origen-SDK/atp/blob/master/lib/atp/flow.rb.
If you were to search for `limit` in there you would come across this line where we can see the structure of a limit node - https://github.com/Origen-SDK/atp/blob/master/lib/atp/flow.rb#L328
So its, [VALUE, RULE, UNIT, SELECTOR].

To get the values, all nodes implement `to_a` (to array), so you could do:

~~~ruby
node   # =>  s(:limit, -2, "gte", nil, nil)
children = node.to_a    # => [-2, "gte", nil, nil]
rule = children[1]
~~~

This lends itself to the following shorthand which you will see in existing handlers:

~~~ruby
node   # =>  s(:limit, -2, "gte", nil, nil)
value, rule, unit, selector = *node
rule  # => "gte"
~~~

Getting back to the handling of the limits, you could either do it within a sub_test handler like this:

~~~ruby
def on_sub_test(node)
  # Some logic to create the current line
  line   # => <current line>
  line.test_name = node.find(:object).value
  node.find_all(:limit).each do |limit_node|
    value, rule, unit, selector = *node    
    case rule
    when "gte"
      line.gte = value
    # ...
  end
end
~~~

Alternatively, here is a two handler solution:

~~~ruby
def on_sub_test(node)
  # Some logic to create the current line
  line   # => <current line>
  line.test_name = node.find(:object).value
  # Assign to an instance variable so the downstream handlers can access this
  @current_line
  # Continue to process the contents of this sub_test node
  process_all(node.children)
end

def on_limit
  value, rule, unit, selector = *node    
  case rule
  when "gte"
    @current_line.gte = value
  # ...
end
~~~

Each have their pros and cons, but certainly the approach of using more handlers generally leads to simpler handlers which are easier to maintain and understand, and is really the way that the AST is supposed to be used. i.e. it is a scalable approach to transforming complex AST data.

Hopefully that will help you to get started, just let me know if there are follow up questions.

pderouen:
@ginty @priyavadan have another look and see if this is ok.

PR #123, opened 2019-06-13
Pattern reader ast

coreyeng:
Hello,

I've had a long running branch to update the decompiler to use an abstract syntax tree and update it to not read in all the vectors at once. This is a second attempt at the decompiler, implemented within those constraints.

I tried to set up as much as possible in a generic way and allow other platforms to reuse what was added, especially for vector based platforms. This includes a parent `Decompiler::Pattern` object, a bunch of base methods, and some base grammars and nodes. There's in README in the decompiler directory containing some notes for adding new decompilers but doubles as a some implementation notes for the existing ones.

In addition to all the base stuff, I've got implementations for J750/Uflex (these two currently share), and the V93K. Right now, the support is pretty basic, but it should be a good starting point. I've got a full specification/regression driver as well. I wanted something to guarantee some consistency and safeguard against changes to the API, but also have something reusable for both additional platforms and enhancements to the existing platforms. I've got some notes on using the specification driver.

I've got user guides as well, so I won't go into any of the actual API here (unless asked). But, at a high level, the decompiler supports what OrigenStil supported, namely: `#add_pins` and `#execute`. Execution support is pretty basic at the moment and really only handles vectors and comments, but opcode/instructions could be added in the future as well. As far as docs go, since I write informally, I tend to get kinda wordy, so any feedback on the docs is appreciated.

If this gets approved, I can look into supporting STIL as well. I'm not familiar with STIL though, so I'd like to make sure this goes through before I sign up to support it.

I also had an issue open for timesets implicitly supporting period settings. In response to [this issue](https://github.com/Origen-SDK/origen/issues/307) this PR also includes some adjustments to the timesets.

The main addition is to support a default `period_in_ns` defined at the DUT level, then being able to call `#set_timeset` with just the name argument and have the `period_in_ns` looked up behind the scenes. This'll allow for knowledge of the timing when only given a timeset name, such as when executing a decompiled pattern.

I wrote a full specification for this and updated the docs appropriately. As I was doing this though, I kept having trouble with the timesets of the tester and the DUT falling out of sync and kept developing circular calls to fix it. So, I pretty much redid the timeset hierarchy. However, all the current specs and regressions pass, except for one, which was edited as the first implementation of the timesets didn't allow the `period_in_ns` to be changed after the tester was cycled. This caused some failures to the Origen examples though, so I took it out for the time being. It could be reinstated easily. The code, docs, and specs are there, just commented out but it would require an Origen update as well.

To isolate to just this plugin, a fair amount of monkey patching went in to this. Long term, this should probably be moved into Origen (or the DUT timesets moved in OrigenTesters), but since prototyping it is kind of a chicken and egg problem, I did so here since I was opening the decompiler PR anyway.


coreyeng:
This originally was added so we could take a pattern generated from Windows and run it through OrigenSim and Cadence, which is only setup for us in Linux. There's more elegant solutions to this but the simplest was just throw the pattern source over the OS wall and simulate it there. Our internal tool prior to OrigenSim supported this, Origen did not.

This feature opens up running and converting sources not generated from Origen, the big one being scan. Currently, if DFT asked us to run a scan pattern, we couldn't. This fixes that. Although writing a simple parser to grab the repeat and vector states is pretty easy, the goal here was provide an official API and a backend structure that allows for future expansion and enhancements. Hence why grammars were used instead of just a few regexes.

I'm not trying to say (sorry if it came off this way) that this is a vast improvement over the existing work flow... Because it's not at all. But it is a missing feature our internal tool supports that Origen doesn't and IMO thats a hole in Origen that needs to be filled.

info-rchitect:
@coreyeng thanks for explaining it to me.

@arkohler Thoughts?

chrisnappi:
Just want to highlight one word in Corey's explanation - 'converting'.  While originally mainly for OrigenSim, this is a big piece needed to be able to enable STIL->tester or tester->tester.  Obviously not Origen's primary function - but good to have as side functionality.

ginty:
Hi @coreyeng,

First off, congrats on completing this, obviously this is a significant piece of work and I really appreciate the effort that has gone into making not just a J750/V93K solution, but also adding a platform for future expansion. And loads of new documentation too :-)

To further the answers already given to @info-rchitect, this really opens up access to a lot of the Origen pattern infrastructure that was previously only available to those who authored patterns natively in Origen, specifically:

1) Origen as a traditional pattern converter, this kind of thing will work globally with no app required:

~~~text
origen convert my_pat.stil -e v93k_smt7       # STIL to SMT7 conversion
origen convert my_pat.avc -e v93k_smt8        # SMT7 to SMT8 conversion
~~~

2) Origen as a virtual tester simulator, basically you just replace the target backend with an OrigenSim tester driver and point to an OrigenSim-compatible compiled design instead:

~~~text
origen convert my_pat.avc -e sim --dut path/to/my/design/INCA_LIBS
~~~

Additionally, for those who do write patterns natively in Origen, this supplements that by allowing snippets written in 3rd party formats to be embedded in your Origen code:

~~~ruby
def startup(options = {})
  # Do some stuff using regular Origen APIs
  pin(:reset).drive!(0)

  # Run mode entry snippet that I got from verification in 93K format
  OrigenTesters::V93K.execute("path/to/mode_entry.avc")

  # Drop straight back into Origen APIs
  my_reg.my_bit.read!(1)
end
~~~

Getting back to the PR itself, architecturally this is much better than the original implementation. As discussed offline @coreyeng, my main concern is whether the penalty for parsing each vector into an AST is worth it or not.

So, what I would like to see is a shoot out between the STIL converter (regex-based) and this new J750 parser which is AST-based.
What's the time for them both to convert a 1M vector pattern into a V93K avc?

Can you enable your new parsers into the compile command? It should be really simple, just a case of updating this convert method - https://github.com/Origen-SDK/origen_testers/blob/master/lib/commands/convert.rb#L69

Try and automatically identify the input format if you can (I like it when Origen is cute like that), and then just return a function that will invoke your APIs to add the pins and execute.

Then, this should be working:

~~~text
origen convert my_pat.stil -e v93k_smt7

origen convert my_pat.atp -e v93k_smt7
~~~

To get a STIL input pattern, the OrigenTesters::STIL driver should be working to create one.

Will be very interesting to get a feeling for what performance we are dealing with here. If they are similar, then the AST approach is definitely able to scale to more complex vector scenarios.
If its much slower though, then maybe that's not worth it since its likely that a regex is good enough - vectors are pretty regular after all.

Another great PR though @coreyeng, well done!

coreyeng:
@ginty, thanks for the feedback! I'll put together some performance monitoring, post the results here, and we can go from there.

coreyeng:
Hello again,

I ran some really simple performance analysis (just using the `time` command) and, as suspected, there's quite performance hit command to STIL. For the test cases I used to develop this (the `workout` patterns), things are fairly close:

`origen analyze_decomp_perf -c 300`
image

And remains so until we get closer to @ginty's 1M pattern. With 1000 vectors (0.1 MiB .atp):

`origen analyze_decomp_perf -c 1000`
image

And for 5000 vectors (0.5 MiB .atp):

`origen analyze_decomp_perf -c 5000`
image

Once we get to the 1 MiB .atp, its hitting an order of magnitude slower:

`origen analyze_decomp_perf -c 10000`
image

For our apps, we ramload a static library (only a few times, after PORs), but the initial load is ~35k vector `.avc` on my current project. Converting that:

`origen analyze_decomp_perf -c 35000`
image

I think about the largest pattern I've seen is ~100k vectors for some scan patterns (I deal with pretty small devices). Converting one of those:

`origen analyze_decomp_perf -c 100000`
image

Here, STIL is a massive 4265 times faster than the J750.

A few notes about this:

* The size of STIL patterns compared to the V93K and J750 are quite a bit smaller. Though, I don't think that plays much of a role in the performance difference.
* The pattern is very simple: it just toggles single pin. I'm assuming that throwing in several pins won't be that much different.
* The memory usage scales much slower this time around. A pattern that's 100x bigger only uses ~2x the memory.
* Unfortunately, it seems like the AST approach does scale exponentially with the vector count:

image

I built this into a command. and check it in. The default count is 10k (corresponds to an almost exactly 1 MiB .atp pattern) but there's options to override and/or scale it, and scale each platform independently. See `origen analyze_decomp_perf -h` for the help message.

ginty:
Hi @coreyeng,

Thanks for following up on this.

Am I reading this right, that to convert a 100K vector STIL pattern only takes < 2s?

I imagine that compares pretty well to the incumbent tool.

No contest for me really about which way to go then here, what are your thoughts?

ginty:
I think there's something wrong, 1M vectors also takes 4s with STIL, and looking at the converted pattern, I see it is empty!
Investigating...

coreyeng:
Same... I just noticed that too. This is what I get for doing this during a meeting... :p

ginty:
Looks like the STIL reader/regex doesn't like the "" around the pattern name and the pin name.

Looks like it might be in the same ballpark, for the 10K pattern I'm now getting 7.5s for the STIL conversion.

ginty:
Just checked in a fix, STIL is still pretty quick be comparison, here for 100K:

~~~text
[INFO]       234.544[0.000]  || Usage Report:
[INFO]       234.544[0.000]  || Count Scales:
[INFO]       234.544[0.000]  ||   j750: 1.0 (100000.0)
[INFO]       234.544[0.000]  ||   v93k: 1.0 (100000.0)
[INFO]       234.544[0.000]  ||   stil: 1.0 (100000.0)
[INFO]       234.544[0.000]  || File sizes:
[INFO]       234.545[0.000]  ||   /home/stephen/Code/github/origen_testers/output/j750/pin_flip.atp: 10.11 MiB
[INFO]       234.545[0.000]  ||   /home/stephen/Code/github/origen_testers/output/v93k/pin_flip.avc: 6.10 MiB
[INFO]       234.545[0.000]  ||   /home/stephen/Code/github/origen_testers/output/stil/pin_flip.stil: 2.10 MiB
[INFO]       234.545[0.000]  ||   User Time:
[INFO]       234.545[0.000]  ||     Max:  101.66 (j750)
[INFO]       234.545[0.000]  ||     Min:  13.1 (stil)
[INFO]       234.545[0.000]  ||     Diff Max-to-Min: 88.56
[INFO]       234.545[0.000]  ||     Scale Max-to-Min: 12.89%
[INFO]       234.545[0.000]  ||     Scale Min-to-Max: 776.03%
[INFO]       234.545[0.000]  ||   System time:
[INFO]       234.545[0.000]  ||     Max:  0.2 (j750)
[INFO]       234.545[0.000]  ||     Min:  0.13 (stil)
[INFO]       234.545[0.000]  ||     Diff Max-to-Min: 0.07
[INFO]       234.545[0.000]  ||     Scale Max-to-Min: 65.00%
[INFO]       234.545[0.000]  ||     Scale Min-to-Max: 153.85%
[INFO]       234.545[0.000]  ||   Elapsed Time:
[INFO]       234.545[0.000]  ||     Max:  102.0 (j750)
[INFO]       234.545[0.000]  ||     Min:  13.0 (stil)
[INFO]       234.545[0.000]  ||     Diff Max-to-Min: 89.00
[INFO]       234.545[0.000]  ||     Scale Max-to-Min: 12.75%
[INFO]       234.545[0.000]  ||     Scale Min-to-Max: 784.62%
[INFO]       234.545[0.000]  ||   Peak Memory Usage:
[INFO]       234.545[0.000]  ||     Max:  129684.0 (v93k)
[INFO]       234.545[0.000]  ||     Min:  113924.0 (j750)
[INFO]       234.545[0.000]  ||     Diff Max-to-Min: 15760.00
[INFO]       234.545[0.000]  ||     Scale Max-to-Min: 87.85%
[INFO]       234.545[0.000]  ||     Scale Min-to-Max: 113.83%
~~~

ginty:
FYI, on this setup it was 132s to convert the 1M vector STIL pattern, so it scales very linearly (13s for 10K).

ginty:
I added a 100pins target, the difference starts to become really apparent then, for 100K vectors and 100 pins, AST = 16.5 mins, Regex/STIL < 1 min:

~~~text
[INFO]       2015.338[0.000] || Usage Report:
[INFO]       2015.338[0.000] || Count Scales:
[INFO]       2015.338[0.000] ||   j750: 1.0 (100000.0)
[INFO]       2015.338[0.000] ||   v93k: 1.0 (100000.0)
[INFO]       2015.338[0.000] ||   stil: 1.0 (100000.0)
[INFO]       2015.338[0.000] || File sizes:
[INFO]       2015.338[0.000] ||   /home/stephen/Code/github/origen_testers/output/j750/pin_flip.atp: 29.19 MiB
[INFO]       2015.338[0.000] ||   /home/stephen/Code/github/origen_testers/output/v93k/pin_flip.avc: 25.18 MiB
[INFO]       2015.338[0.000] ||   /home/stephen/Code/github/origen_testers/output/stil/pin_flip.stil: 11.65 MiB
[INFO]       2015.338[0.000] ||   User Time:
[INFO]       2015.338[0.000] ||     Max:  990.9 (j750)
[INFO]       2015.338[0.000] ||     Min:  50.53 (stil)
[INFO]       2015.338[0.000] ||     Diff Max-to-Min: 940.37
[INFO]       2015.338[0.000] ||     Scale Max-to-Min: 5.10%
[INFO]       2015.338[0.000] ||     Scale Min-to-Max: 1961.01%
[INFO]       2015.338[0.000] ||   System time:
[INFO]       2015.338[0.000] ||     Max:  0.5 (v93k)
[INFO]       2015.338[0.000] ||     Min:  0.2 (stil)
[INFO]       2015.338[0.000] ||     Diff Max-to-Min: 0.30
[INFO]       2015.338[0.000] ||     Scale Max-to-Min: 40.00%
[INFO]       2015.338[0.000] ||     Scale Min-to-Max: 250.00%
[INFO]       2015.338[0.000] ||   Elapsed Time:
[INFO]       2015.338[0.000] ||     Max:  992.0 (j750)
[INFO]       2015.338[0.000] ||     Min:  50.0 (stil)
[INFO]       2015.338[0.000] ||     Diff Max-to-Min: 942.00
[INFO]       2015.339[0.000] ||     Scale Max-to-Min: 5.04%
[INFO]       2015.339[0.000] ||     Scale Min-to-Max: 1984.00%
[INFO]       2015.339[0.000] ||   Peak Memory Usage:
[INFO]       2015.339[0.000] ||     Max:  148644.0 (stil)
[INFO]       2015.339[0.000] ||     Min:  125360.0 (j750)
[INFO]       2015.339[0.000] ||     Diff Max-to-Min: 23284.00
[INFO]       2015.339[0.000] ||     Scale Max-to-Min: 84.34%
[INFO]       2015.339[0.000] ||     Scale Min-to-Max: 118.57%
~~~

coreyeng:
Thanks for the additional analysis, @ginty! Seems more realistic than my original numbers now. What are your thoughts on the performance difference? I do think that the AST and Treetop grammars can handle more complex input a bit more cleanly, though that may be because I understand the Treetop and AST better now. I would like to keep a lot my 'vector type' stuff. I think that would allow for more complex execution support later on, and helps various platforms have some commonality.

I wonder if the main time sink is using Treetop to parse the vectors. I think STIL just uses a straight regex to do so? I know Treetop has a lot of overhead but I thought it was mostly just startup. Possibly not? Anyway, I wonder if providing our own regex block and manually building our own ASTs would improve the performance sufficiently. Treetop does give us a nice interface to build grammars though.

ginty:
Hi @coreyeng,

My opinion is pretty much un-changed, parsing to an AST at vector-level is overkill and I don't think it can be justified, especially now that it has been confirmed that we are paying an order of magnitude performance hit for it.
I think that "a free pattern converter that can output to SMT8" is a pretty good sales pitch and I expect the addition of this feature to put Origen into many engineer's hands for the first time - so I am really keen that we make a good impression.

I think the performance difference at the smaller end of the scale will actually be the most noticeable - when converting a smaller pattern an engineer is likely to just sit and wait for it to complete rather than go off and do something else. When watching and waiting for it to complete, a minute will feel very different from a few seconds and it is in that minute that an opinion of Origen will be formed.
I would hate to see us shooting ourselves in the foot here over the frontend user experience for the sake of engineering elegance in the backend.

So, my opinion is that the vector processing side of this should be greatly simplified, to something more like how STIL does it, which is just a single regex-based method to parse it - https://github.com/Origen-SDK/origen_stil/blob/master/lib/origen_stil/pattern.rb#L87
And another method to write it - https://github.com/Origen-SDK/origen_stil/blob/master/lib/origen_stil/pattern.rb#L16

You can add more object structure around it if you want, but be aware of the performance, which I guess is easy enough to do now that you've added this nice performance regression test for us.
My feeling is that most of the time is lost in the parsing stage, so you could experiment with parsing via regex and then creating our own AST, but be aware that I will still complain if it is significantly different performance-wise vs. the STIL baseline.
The bottom-line for me is that pattern vectors are very well defined and don't have much variation to deal with, so I think regexes for parsing and pretty simple sequential logic for dispatch are the right tools for this job.

I do feel bad giving this feedback and I really hope it does not discourage you - it is  clear that you have already put a lot quality time and effort into this.
I wouldn't view it as time wasted though, as you say, you understand the AST creation and processing a lot better now and that will pay dividends in many other areas of OrigenTesters and the Origen eco-system where the power that comes from using an AST is justified.

Ultimately, the reason that I am being picky on this point is that you are providing a very high-impact feature here, so let's make sure that we do it right.

chrisnappi:
There may still be some cases where the AST method makes sense.

Perhaps when we are going straight vector to vector (in the format called 'serial' by DFT engineers) using the RegEx makes sense.

But we do also get WGLs is parallel format (though this has been less common as of late) - and I assume that STIL has a similar convention.   This format is much more compact in WGL format, as it has templates for each type of vector (shift/capture) and these templates are then followed by scan data for an entire shift in/shift out on single lines (hence the confusing 'parallel' term).

Either way(regex or AST), I think we should aim to support this format - while we have mostly beaten our DFT engineers into submission to deliver 'serial' format, in the industry my understanding is parallel is used a lot.   This translation definitely requires more processing sophistication.

@info-rchitect @aaronburg - assuming it is not proprietary information :) which format do you receive scan vectors in?

info-rchitect:
@arkohler Can you answer @chrisnappi ?  I think you would provide the best answer from a user perspective.

ginty:
Good point @chrisnappi, and I don't really know what WGL looks like to say whether an AST is needed to process it or not.
For me, you need an AST either if you are processing something which is language-like and the incoming data could be formatted in the infinite number of ways that a user might express something using that language.
For example, I did use an AST to handle the front portion of a STIL pattern, because STIL basically defines a language that can be used to describe timing, pins, etc.
The vector portion of it though is very strictly defined, similar to vectors in a .atp and .avc - I presume because ATE's want to have as little parsing overhead as possible when it comes to playing vectors.
Anyway, no AST required for that.

The other time you might need it would be if the contained data itself contains complex logic to be transformed, the Origen test flows are a good example of that. Transforming those from Ruby code into a Teradyne spreadsheet format or a SMT C-like format is really quite hard to do and the AST gives the ability to do that quite easily and most importantly with the confidence that the transformed version contains the correct logic.
Again, for vectors we don't need that.

For higher-level WGL...maybe.

STIL does have a similar concept, they call cyclized or not patterns. Currently, the STIL importer assumes cyclized STIL patterns.
As far as STIL goes though, I still think a regex is good enough to handle them.

However, you make a compelling case for keeping all Corey's infrastructure around, but I just don't think we should be using it for the cases of consuming .atp and .avc vectors.

chrisnappi:
Stephen,  I don't know STIL very well - but I think the equivalent there is the ScanStructure block.

Good WGL reference: https://www.tessi.com/uploaded/files-download/TSSI_WGL_LanguageGuide.pdf
STIL: http://grouper.ieee.org/groups/1450/dot1/for-discussion-sep23.pdf

Again - I do agree that regex looks to make sense for the general case - but wanted to being up this more complex translation case before throwing out the bath water.

coreyeng:
Hi @ginty, yeah, I thought your opinion wouldn't be changed. I think the current backend has its merits, but I actually agree that the performance hit isn't worth it, especially with it scaling exponentially. I didn't think much about performance, I was more concerned about re-use, future expansion, and flexibility.

I'll start by swapping out Treetop with just some plain regex method for vector parsing and see where that goes. I know you view this for pattern conversion, but my immediate use case is in running though simulation, which I like the AST processors and 'vector types' for, as I think it greatly simplifies handling vectors outside of just a translation from source to source. Hopefully swapping out the vector parser will be sufficient, but we'll see.

As for the discouraging note, everything's been constructive and with sound reasoning, and its gotten me to think in different constraints than usual, so its a good learning experience. As long as there's no deadline, this'll just be a continuing project that I'll slowly but surely make progress on. The big thing for me is the API. Most of the work since probably March (only got a few hours a week, if that, to work on this) went into the regressions and documentation. As long as that seems sound, progress still seems like its being made. And, I can re-use the regressions to verify the backend changes!

ginty:
@chrisnappi, that scan structure is not supported/implemented yet by the STIL parser, but it would be parsed to an AST whenever it is.

The structure of a STIL file is roughly:

~~~text
<language-like block>
<language-like block>
<language-like block>

<pattern block>
  <vector>
  <vector>
  <vector>
  <vector>
  <vector>
</pattern block>
~~~

The upper part is parsed to an AST  (which is where the scan description blocks will go) but the vector portion is not.
Even in this current PR the pattern vector block is not parsed to a complete AST - that would take forever with large patterns, at least with this Ruby parser.
So when it comes to working through the vectors, the AST is really only helping to deal with any horizontal/within-vector complexity, but I don't think that any significant complexity really exists at that level.

I had a quick scan through the WGL spec and it looks pretty similar, probably STIL was heavily influenced by it, and I would propose to handle it in much the same way.

ginty:
@coreyeng, ok glad to hear we are reaching alignment.

On the simulation case, I don't follow why we need to treat that any different from the conversion case?

I was trying VT-like simulation of a STIL this week and it was just a case of swapping a V93K driver (conversion) to a sim-driver (VT).

arkohler:
I found this in case it helps.
STIL BNF guide:  http://grouper.ieee.org/groups/1450/dot0/bnf.pdf

arkohler:
@chrisnappi, we receive our scan vectors in a sort of hybrid format.
We'll have a series of V {} vectors like before and between scan shifts.  
  W scan_cycle;
  V { /* Vec 12 Pat_Cycle 37 Tot_Cycle 37  */
    REFCLK     = PpPp;
    JTAG       = 0011Z;
    SCAN_CTRL  = 010;
    SCAN_IO    = 00000000ZZZZ0000ZZZZ000000000000;
}

The the shifts themselves use a per-pin format with a macro, like this:
  Macro "Scan_Macro_Shift" { /* Vec 13 Pat_Cycle 38 Tot_Cycle 38 */
    PIN_0 = 0000001111000110;
    PIN_1 = 0100110000010000;
    PIN_2 = 1001110000110010;
    PIN_3 = HHHHLHLLLLHHLL;
}

I'm not sure how fast it would be, but with something like [Parselet](https://kschiess.github.io/parslet/documentation.html) and the BNF I mentioned above, you could probably support a much wider variety of valid STIL syntax than RegEx checks would allow.

coreyeng:
Its not required, but I think it helps with the organization. We'll see how much of an impact it makes though. If its still really slow, I'll take that out too.

chrisnappi:
Thanks @arkohler .

@ginty this is the format I was talking about - it is not just header information then vectors, but rather a vector structure then per-pin data.  This is a pretty common format and requires a totally different scheme.  That said, it can be done pretty easily simply slurping the per pin data in to hashes for each scan in/scan out sequence and then rendering.

ginty:
@chrisnappi, that's actually a pretty regular STIL vector, the per-pin aspect of it is often not apparent because its often converted into a structure to define the values for "ALL" pins, but multi-pins is already handled (with a regex) - https://github.com/Origen-SDK/origen_stil/blob/master/lib/origen_stil/pattern.rb#L72

coreyeng:
I'm using this branch internally and need to push some non-refactoring stuff to it. I'll close this now and reopen when I've got something else to review.

PR #126, opened 2019-06-26
Added support for boolean types in V93K test method attributes

ginty:
This allows custom test method definitions for the V93K (SMT7) to declare attributes as boolean type and then assign them true/false and they will be rendered as a 1/0 integer.

This update has been on the SMT8 branch for a long time, but bringing it over to the mainline since its usage has crept into origen_std_lib.

### DIFF
--- a/lib/origen_testers/smartest_based_tester/base/test_method.rb
+++ b/lib/origen_testers/smartest_based_tester/base/test_method.rb
@@ -2,7 +2,7 @@ module OrigenTesters
   module SmartestBasedTester
     class Base
       class TestMethod
-        FORMAT_TYPES = [:current, :voltage, :time, :string, :integer, :double]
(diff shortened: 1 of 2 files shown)


priyavadan:
LGTM!

ginty:
I was just reviewing why the integer and double types are not converted to a string as part of this change.
That makes no difference for SMT7 - downstream when rendering in the template everything is converted a string anyway and all parameter values get surrounded by quotes in the test flow file.

However, on SMT8 the numerical values do not get quoted so the conversion to a string needs to be withheld until later to allow the template to know whether to add quotes or not.

Anyway, I'll leave it in as it will be coming over as part of the SMT8 PR soon enough anyway and the test coverage in here confirms there is no difference to the SMT7 output.

PR #132, opened 2019-08-06
fixes flag comparisons between symbols and string

info-rchitect:
I still need to make an example and perhaps a spec test but this is one of the changes we had to do to use this gem.


info-rchitect:
@priyavadan Can you pls file the PR for the examples bug we found?  I would like to use that fix so adding more examples for this PR goes much quicker.

priyavadan:
> @priyavadan Can you pls file the PR for the examples bug we found? I would like to use that fix so adding more examples for this PR goes much quicker.

@info-rchitect understood, I'll probably start another PR for that as it is just a cosmetic issue and doesn't effect generation; IMO

priyavadan:
@info-rchitect  looks good to me 

@rlaj any concerns we merge this soon?

PR #134, opened 2019-08-15
enabled test interface to override default behavior of reloading targ…

info-rchitect:
…ets no matter what

### DIFF
--- a/lib/origen_testers/origen_ext/generator/flow.rb
+++ b/lib/origen_testers/origen_ext/generator/flow.rb
@@ -92,6 +92,9 @@ module Origen
 
       # @api private
       def _create(options = {}, &block)
+        options = {
+          reload_target: true
+        }.merge(options)
         # Refresh the target to start all settings from scratch each time
         # This is an easy way to reset all registered values
         Origen.tester.generating = :program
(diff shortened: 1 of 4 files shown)


priyavadan:
@info-rchitect can you please look into why the CI builds are failing. Seems like the flow_control test flow has a comment difference

info-rchitect:
The last commit was just code style to ensure the private method returns a Boolean in all situations

priyavadan:
@ThaoHuynhFsl this should solve the long generation times for ASF as well, I think; do you want to trial it out please?

lib/origen_testers/origen_ext/generator/flow.rb
```diff
@@ -106,6 +108,10 @@ def _create(options = {}, &block)
         job.output_file_body = options.delete(:name).to_s if options[:name]
         if sub_flow
           interface = Origen.interface
+          if reload_target?(interface, options)
+            Origen.app.reload_target!
+            Origen.tester.generating = :program
+          end
```
ginty: Can you comment on why the reloading has been moved from its original position?

info-rchitect: @ginty Previously the target was reloaded before the interface was instantiated so there was no chance for the interface to alter the outcome.

info-rchitect:
> OK, makes sense, thanks for the clarification.
> I think this feature should be briefly mentioned in either the Flow or Interface guide though.

@ginty will do, one sec.

PR #135, opened 2019-08-18
Pattern reader regex

coreyeng:
Third time 's a charm??? The last PR had major execution time concerns, most of which seems to have been due to Treetop/AST. I stripped that out completely and now we have some fairly simple `regex` parsers. The execution times are greatly improved, even with the user interface intact, and really the only thing lost are syntax error handling. Though, I think this is minor (for the time being) since @ginty's main usage is pattern conversion and mine is running sources through `OrigenSim`, both of which will have sources which should be genuine and syntactically correct. I do have some 'catch alls' that, if an exception is raised, will print out which vector the error occurred on and what the exception was. So, not turning a blind eye to syntax errors, but not quite the 'expected ...' that Treetop would have given us.

As previously said, I have the same interface as before. The run times didn't seem overly affected by this, at least not until we hit 100k+ vectors, and even then much less so than before. All the docs should still be relevant, as these should just be backend changes and I updated the README in `lib/origen_testers/decompiler` to account for these changes. I'll make another pass through the documentation, but I wanted to open this so we can review the execution times and code. The previous specs pass with very minor changes, mostly just error messages resulting from the less-specific/more Ruby-centric parsing errors than what Treetop had.

I added a `ramp` option to the built-in performance monitor. This'll do the same thing as before, but ramp from 1k vectors up to 1 million.

So, here's the new performance:

image

image

Base   Count (vectors) | j750 (User Time (Seconds)) | v93k (User Time (Seconds)) | stil (User Time (Seconds))
-- | -- | -- | --
1000 | 1.02 | 1.11 | 1.07
5000 | 1.6 | 2.1 | 1.41
10000 | 2.33 | 3.37 | 1.86
50000 | 8.29 | 13.51 | 5.69
100000 | 16.41 | 27.73 | 10.64
500000 | 81.94 | 140.57 | 50.75
1000000 | 175.12 | 291.85 | 104.58

I didn't do any automatic averaging though, just in my own usages, I haven't seen the run times differ by more than 5%, even at a million vectors.

A noticeable difference won't really be seen between the platforms until somewhere around 10k to 50k vectors, I still like the idea of keeping the user interface intact, as it at least has the illusion of some commonality and consistency between the supported platforms. Additionally, looking at the difference in run times between the J750 and V93K, both of which support the interface, the time difference could just be the nature of the V93K, or I could just have poorly written `regexes` for it and need to optimize.

Finally, just as a reminder (since its been a while), this also contains the `timeset` changes that I had in the original from almost a year ago. I've been using it in my apps for pretty much this entire time and it seems fine. All the existing specs/examples still pass, and this feature itself is also documented and spec-ed.

Thanks!


redxeth:
Given the project using this branch currently we had a very minor change needed- so added option to skip digcap AIV setup here to this branch that that project needed.

ginty:
A bit sneaky @redxeth, correct flow would have been to branch from here and add your changes in its own PR to here.

ginty:
Hi @coreyeng, looking good now performance-wise, I have no issue with some relative differences between the formats now that the intrinsic order-of-magnitude difference has been resolved.

Not sure how easy it would be for you to run the performance tests using - https://github.com/Origen-SDK/origen_stil/pull/1
I had to add quite a bit more complexity to the STIL vector parsing and I would be surprised if it is still so much of a frontrunner now anyway.

The only slight reservation I have is with the documentation.
As ever, well done for putting the effort in to write it, but I'm wondering if this is really user/app developer documentation that belongs in the guides?
It seems more like real power-user stuff to me. We already have the advanced section, but since you have added multiple pages it makes sense to give it its own section.

I'm wondering if we should start a whole other doc area, 'Developer Guides' or something?

We already have a few other pieces of similar documentation out there looking for a home, e.g. the OrigenVerilog README, which is also not in the guides because I felt that it was just noise that the average user should not be concerned about.

For de-compilation, I think the average user should not need to know about it. They should know about the convert command and similar higher-level things built upon it. Those things should be in the main guides, the lower level infrastructure should probably be somewhere else.

What do you think?

Great work though, no issues with merging from a functionality POV.

coreyeng:
Shouldn't be too much trouble. I've got everything saved. Will point the Gemfile to that commit and give it a go.

Yeah, the docs are more in-depth than the average user needs. I tried to keep the overview page as simple as I could. I'm still going for a single interface to fit all patterns though and, although you're right, it'll be noise for most users, I wasn't quite sure where to put it. My use case goes beyond just converting too. I'm still striving to support as much on the simulation side as I can so I actually use more than just the `origen convert` command and putting all the extras in the docs/specs make it more official, to me at least.

I think the advanced section might be a better place for it? I wasn't sure if I can inject pages into other sections though and since it is part of the decompiler, I left it where it is.

coreyeng:
As for STIL, didn't see to make much of a difference. Rerunning shows everything going down by ~0.1 seconds, though I guess that's part of problem with just using the `time` command.

image

image


Base Count   (vectors) | j750 (User Time (Seconds)) | v93k (User Time (Seconds)) | stil (User Time (Seconds))
-- | -- | -- | --
1000 | 0.97 | 1.1 | 1
5000 | 1.5 | 2 | 1.31
10000 | 2.21 | 3.18 | 1.77
50000 | 7.78 | 13.3 | 5.38
100000 | 15.57 | 26.05 | 9.91
500000 | 79.22 | 136.42 | 49.12
1000000 | 167.27 | 272.03 | 101.23

ginty:
OK that's good, thanks.
Thought it might be worse, by happy to see I still beat you :-)

coreyeng:
Haha. Fair enough.

Do you want to handle the docs now, or merge and see about shifting stuff around when we get a developer/more advanced section in place?

redxeth:
> A bit sneaky @redxeth, correct flow would have been to branch from here and add your changes in its own PR to here.

I spoke with @coreyeng and he was ok with it given the project this branch is using.

coreyeng:
@ginty, are you okay with a merge here then? And we'll shift around the docs a later time?

PR #136, opened 2019-08-30
Added option to skip digcap AIV setup.

redxeth:
Minor update to permit skipping digcap setup in AIV for v93k.


redxeth:
verbal approval given

PR #137, opened 2019-10-02
DRY flow works but not sure if this is the correct implementation

info-rchitect:
Will not work for SMT7 as currently architected due to the fact that SMT7 does not store the lines to be printed to the ERB template in an object.  Rather it finds the test method parameter values in the [ERB template](https://github.com/Origen-SDK/origen_testers/blob/master/lib/origen_testers/smartest_based_tester/v93k/templates/template.tf.erb#L140-L145).  Whereas the SMT8 implementation keeps the flow file lines stored as part of the [test suite instance](https://github.com/Origen-SDK/origen_testers/blob/master/lib/origen_testers/smartest_based_tester/v93k_smt8/test_suite.rb#L53-L59).

Here is how I added the programmed parameters in my test interface:

~~~ruby
      test_suite_opts = test_instance.primaries.merge(test_instance.flags).merge( { programmed_params: test_instance.test_method_params.ids })
      t = test_suites.add(test_instance.test_suite_name, test_suite_opts)
~~~

### DIFF
--- a/lib/origen_testers/smartest_based_tester/base/test_suite.rb
+++ b/lib/origen_testers/smartest_based_tester/base/test_suite.rb
@@ -2,10 +2,18 @@ module OrigenTesters
   module SmartestBasedTester
     class Base
       class TestSuite
-        attr_accessor :meta
+        attr_accessor :meta, :programmed_parameters
 
         def initialize(name, attrs = {})
           @name = name
+          unless attrs[:programmed_params].nil?
+            unless attrs[:programmed_params].is_a? Array
+              Origen.log.error('Test suite user programmed parameters must be passed as an Array of test method parameters')
+              fail
+            end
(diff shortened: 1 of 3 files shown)


ginty:
Can you comment on what the intention of this change is?
e.g. an example of the change in the output you want to see
Thanks!

info-rchitect:
> Can you comment on what the intention of this change is?
> e.g. an example of the change in the output you want to see
> Thanks!

Unprogrammed test method parameters are not printed to the generated flow file, reducing file size and improving readability.  For SMT8 it also prevents overwriting default values within the test method whereas SMT7 it was OK to program an empty String.

info-rchitect:
@ginty can we close this and figure out a backwards compatible solution for SMT7?

PR #139, opened 2019-11-05
pull request for minor update to allow import of top level flow files during test program generation or flow simulation

jiangliu23:


lib/origen_testers/generator.rb
```diff
@@ -254,8 +254,16 @@ def reference_file
     end
 
     def import(file, options = {})
+      options = {
+        component: true # by default import method treats file as a component flow file e.g. _my_component.rb
+      }.merge(options)
```
redxeth: I think this goes against what @ginty had in mind-- that it should just work whether the user is importing a component or a sub-flow.  Here it is required that user know to set `component: false` when using `import` for a sub-flow. 

I think to provide the "it just works" functionality you would have to check to see whether file exists or not using the filename as-is but WITH the underscore.  If not, then check it also without the underscore.  This way if there are 2 files, one with underscore and one without, the underscore one would take priority.  

With that, then no passing of option is needed.

ginty: yes, agreed

jiangliu23: okay i misunderstood i think, thanks

priyavadan:
Just thinking out loud here...

Could one not overload(correct term?) the import method in the users own interface and pass it to this method via 'super'?

In the local interface something like below:
 
def import (options)
  # code here to check if component file or not...
  super
end

redxeth:
I think we're looking for a solution more people can share.  I suppose it could have been done in the local interface but I think he's looking for something to share across projects.

lib/origen_testers/generator.rb
jiangliu23: @ginty @redxeth , let me know if this makes sense, i've upshed in some new changes.

priyavadan:
CI checks are failing, can those be looked at please? thanks!

`Can't find: /home/travis/build/Origen-SDK/origen_testers/program/_custom_tests.rb`

jiangliu23:
> I don't think this works right, e.g. if you don't append the ".rb" when importing a top-level flow it will not work and will fall through to the regular component lookup.
> 
> Generally, I don't like the direction here of duplicating parts of `clean_path_to_sub_program`. At this point in the code the change should be very simple:
> 
> ```ruby
> # Try a regular component name first with leading _, then try for without
> file = Origen.file_handler.clean_path_to_sub_program(file) || 
>        Origen.file_handler.clean_path_to_sub_program(file, with_underscore: false)
> ```
> 
> A corresponding update is then required to add the with_underscore option handling in the existing lookup method.
> 
> Thanks for the effort so far though.



> I don't think this works right, e.g. if you don't append the ".rb" when importing a top-level flow it will not work and will fall through to the regular component lookup.
> 
> Generally, I don't like the direction here of duplicating parts of `clean_path_to_sub_program`. At this point in the code the change should be very simple:
> 
> ```ruby
> # Try a regular component name first with leading _, then try for without
> file = Origen.file_handler.clean_path_to_sub_program(file) || 
>        Origen.file_handler.clean_path_to_sub_program(file, with_underscore: false)
> ```
> 
> A corresponding update is then required to add the with_underscore option handling in the existing lookup method.
> 
> Thanks for the effort so far though.

So from the perspective this gem itself, how would it know which version of Origen Core it needs to pull to meet the minimum dependency requirement?

jiangliu23:
it's been modified according to the exact recommendation, please review and see if this meets the requirement.

jiangliu23:
Okay thanks.

redxeth:
Am adding the capability on the origen side-- but it will require a couple of additional updates on the origen_testers side as well.

Namely since we are doing an or `||` then we need to allow for one of them to not find the file without error.  That's been handled but it means we have to allow also for if neither are found, so a check for `file.nil?` needs to be added to induce a failure if file not found.

Second, even if the import is working properly, we cannot use the example given with importing `custom_tests.rb`.  The reason is that that flow file uses a custom interface defined in its `Flow.create` header, which is lost when imported so standard interface has no idea what `custom` meanas.

This could be why the CI is failing.  The CI is likely trying to generate the `prod.list` and having this issue.  I would recommend a different subflow to import that uses the same interface as that which is importing it.

Also, I'm not sure yet but if we modify any of the flows, likely we will also need to modify the approved files that they are compared against.  We'll see once we fix the first things above.

redxeth:
Ok-- pull request started in Origen: https://github.com/Origen-SDK/origen/pull/364.

Code to update in origen_testers is:

In `generator.rb`:
~~~ruby
 def import(file, options = {})
      file = Pathname.new(file).absolute? ? file : "#{current_dir}/#{file}"
      filemsg = "Could not find file to import: #{file}"
      # Try a regular component name first with leading _, then try for without
      file = Origen.file_handler.clean_path_to_sub_program(file) || Origen.file_handler.clean_path_to_sub_program(file, with_underscore: false)
      fail filemsg if file.nil?  # give error if neither option above is found
    ...
~~~

and in `prb1.rb` instead of line added before:
~~~ruby
 import 'test' # import top-level test.rb directly
~~~

Travis CI may still fail-- you'll have to see how to update what the program generation compares against.  Or just put comment in front of the example import statement and say it's an example.

redxeth:
From talking to @jiangliu23 and @ginty, I'll go ahead and start my own branch to support these types of changes and start a separate PR. Once that happens we can just close this one.

redxeth:
Can close this PR as https://github.com/Origen-SDK/origen_testers/pull/140 replaces it.

PR #140, opened 2019-11-08
Allow import of top-level flow during test program generation

redxeth:
This PR to replace https://github.com/Origen-SDK/origen_testers/pull/139.

Basically want to be able to use the import statement in a program Flow to be able to not only import a flow component but also a top-level flow.   As before the user provides the sub-flow or component name without an underscore but now the import method will look for either.  If both files exist then the one with the underscore takes priority.

So basically if you have these files in the flow directory:

    _mycomponent.rb
    mysubflow.rb
    _myambiguousfilename.rb
    myambiguousfilename.rb

and do this flow:
~~~ruby
Flow.create do

   import "mycomponent"             # will import _mycomponent.rb
   import "mysubflow"               # will import mysubflow.rb
   import "myambiguousfilename"     # will import _myambiguousfilename.rb

end
~~~

If it cannot find either file, it will give an error:

    Could not find file to import: <full path to last file tried to find>"

In addition, the dependency on the Origen file_handler method `clean_path_to_sub_program` has been removed, so now that method can be removed from Origen also, as this was the only place that used it.


info-rchitect:
LGTM, will docs be updated?

redxeth:
All of the related documentation is under Origen, not Origen Testers, unfortunately. 

image

I think bringing all of the documentation over from there is what may be warranted but is a bigger thing than I wanted to handle for this PR. 

https://origen-sdk.org/origen//guides/program/doc/?highlight=import%20command

redxeth:
Sorry, @ginty explained to me how the OrigenTesters plugin doc templates will be eventually included in the next origen release (built from origen_testers plugin).  So will try to update docs a bit for this PR.

redxeth:
Docs updated, please take a look.

PR #141, opened 2019-12-03
Initial checkin for group bypass on the 93K

priyavadan:
Before I add in the documentation for this, @info-rchitect can you please try this out and let me know if this satisfies your use case? Thanks!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gem 'yard-activesupport-concern'
 gem 'origen_jtag', '>= 0.12.0'
 gem 'origen_doc_helpers'
 gem 'nokogiri', '1.8.5'  # Lock to the version to enable testing in Ruby 2.2
+gem 'dry-inflector', '0.1.2'
 
 gem 'origen_stil', git: "https://github.com/Origen-SDK/origen_stil.git"
 
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gem 'yard-activesupport-concern'
 gem 'origen_jtag', '>= 0.12.0'
 gem 'origen_doc_helpers'
 gem 'nokogiri', '1.8.5'  # Lock to the version to enable testing in Ruby 2.2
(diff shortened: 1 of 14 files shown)


priyavadan:
@ginty looks like dry-inflection 0.2.0 can't run on ruby < 2.4.0 , not sure yet what is calling that gem

priyavadan:
@info-rchitect okay cool

@ginty @info-rchitect should we push in that here as well? probably so; to avoid conflicts in user environments

info-rchitect:
@priyavadan yes might as well

priyavadan:
great, will add documentation tomorrow and we can then merge if all okay! Thanks!

Gemfile
```diff
@@ -9,6 +9,7 @@ gem 'yard-activesupport-concern'
 gem 'origen_jtag', '>= 0.12.0'
 gem 'origen_doc_helpers'
 gem 'nokogiri', '1.8.5'  # Lock to the version to enable testing in Ruby 2.2
+gem 'dry-inflector', '0.1.2'
```
ginty: Can you add a comment as to why this is here so we don't forget

lib/origen_testers/smartest_based_tester/base/flow.rb
```diff
@@ -586,12 +590,18 @@ def on_group(node)
           stack[:on_fail].pop if on_fail
           stack[:on_pass].pop if on_pass
           @indent -= 1
+          bypass = node.find(:bypass).try(:value) || flow_bypass || false
```
ginty: The false at the end is redundant here

Gemfile
priyavadan: added

lib/origen_testers/smartest_based_tester/base/flow.rb
priyavadan: updated!

priyavadan:
> LGTM, couple of minor housekeeping things noted in comments.
> Thanks!

updated per feedback!

PR #144, opened 2020-01-15
Sub flow name as an option to the import method

info-rchitect:


info-rchitect:
@ginty do you know why the travis builds are still failing after I updated the rubyzip version in the gemspec?

ginty:
Looks good.

I think it needs a tweak to put nils in the stack. I think the way it is currently it will not work if a sub_flow put a name in the stack and it then included an embedded sub-flow with the default name. In that case because `stack.empty?` will be false the embedded flow will end up getting the parent's name too.

So it should always put either nil or a name on the stack and then `stack.last` will always return the name variable for the given flow.

lib/origen_testers/origen_ext/generator/flow.rb
```diff
@@ -12,7 +12,8 @@ class Flow
       # of the ATP AST originated from
       def create(options = {}, &block)
         options = {
-          reload_target: true
+          reload_target: true,
+          name:          OrigenTesters::Flow.name_stack.empty? ? nil : OrigenTesters::Flow.name_stack.last
```
ginty: I don't think we need this empty? check anymore, maybe we should just pop the stack here.

origen_testers.gemspec
```diff
@@ -34,4 +34,5 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'origen_stil', '>= 0.2.1'
   spec.add_runtime_dependency "ast", "~> 2"
   spec.add_runtime_dependency "sexpistol", "~> 0.0"
+  spec.add_runtime_dependency 'rubyzip', '1.3.0'
```
ginty: Can you remove this, I've pushed an update to master which should hopefully fix it

info-rchitect:
@ginty thx for the travis help.  However I don't see any jobs being started now.

ginty:
Ok, master is building again - https://travis-ci.org/Origen-SDK/origen_testers/builds/637510338?utm_medium=notification&utm_source=github_status

Can you merge in latest master here please and hopefully it will work.

info-rchitect:
@ginty I merged master locally but how do I get the branch to see it as changes and re-run travis builds?

info-rchitect:
thx @ginty

PR #148, opened 2020-02-24
Inhibit vectors bugs

ginty:
Came across a corner case bug where a match loop called within an inhibit vectors block would crash if no vectors had been previously generated.

I added more test cases here to run the workout patterns fully inhibited and came across a few more similar cases.


ginty:
Also dropped testing for Ruby 2.3, seems like the Ruby community has moved on and so we should too.

PR #150, opened 2020-03-13
bugfix for force_serial implementation

priyavadan:
@info-rchitect @ginty I believe this is the correct implementation for force_serial local_flag.


info-rchitect:
@priyavadan Why is the travis build still in progress?

priyavadan:
@ info-rchitect no clue but the results show that the builds finished and passed.

priyavadan:
Also waiting on internal verification of the implementation before merging.

PR #151, opened 2020-03-30
Skip dig src start

pderouen:
There are a few rare scenarios when you don't want Origen to automatically insert the digsrc instrument start command. This PR provides a way to turn it off.

### DIFF
--- a/approved/generic_overlay_capture/ultraflex/test_overlay_no_start.atp
+++ b/approved/generic_overlay_capture/ultraflex/test_overlay_no_start.atp
@@ -1,0 +1,123 @@
+// ***************************************************************************
+// GENERATED:
+//   Time:    30-Mar-2020 14:55PM
+//   By:      Paul DeRouen
+//   Mode:    debug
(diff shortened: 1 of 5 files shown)


ginty:
Looks good.
Could we add a quick note on this to the (currently still a placeholder) UFlex pattern guide? - https://github.com/Origen-SDK/origen_testers/blob/master/templates/origen_guides/pattern/ultraflex.md.erb

pderouen:
Is there a way to compile/serve the template to make sure it's what I expect?

pderouen:
I just pointed a local origen workspace's Gemfile to my local testers gem and built the origen web docs to view it.

Ok, I think this should do the trick.

ginty:
> Is there a way to compile/serve the template to make sure it's what I expect?

Yeah, it should just be `origen web compile templates/origen_guides` from within this app.

Thanks for the doc addition!

pderouen:
> 
> 
> > Is there a way to compile/serve the template to make sure it's what I expect?
> 
> Yeah, it should just be `origen web compile templates/origen_guides` from within this app.

LOL, should've been obvious to me! Thanks.

Issue #70, opened 2017-10-16
V93k tmm_console utility useful for creating test method libraries automatically


info-rchitect:
Hi,

There is a command on the V93K tester tmm_console tool:

~~~shell
/opt/hp93000/soc/testmethod/bin/tmm_console -e dump types
~~~

If you redirect it to an ASCII file you will get a file that looks like this:

~~~text
testmethod 1 -- ccf_tml.CCT_SYNC
  location name: "CCT_SYNC.cpp:CCT_SYNC"
  shared lib path: "/opt/hp93000/soc/segments/tml/sh_lib/preload/libccf_tml.so"
  source project path: ""
  compiled with SmarTest version: "7.3.2.11, 05-Jul-16"
  parameter ids:
    CCF_ContextType :string, Synchronized
    CCF_FocusedPins :ContextPins, 
    CCF_AnalogRFSetName :string, 
    mIsRetest :string, 
  limit ids:
  testnumber ids:
testmethod 2 -- ccf_tml.CCRetest
  location name: "CCRetest.cpp:CCRetest"
  shared lib path: "/opt/hp93000/soc/segments/tml/sh_lib/preload/libccf_tml.so"
  source project path: ""
  compiled with SmarTest version: "7.3.2.11, 05-Jul-16"
  parameter ids:
    maxRepeatCountOn :string, OFF
    maxRepeatCount :int, 0
    SMC_SYNC_ON :string, OFF
  limit ids:
  testnumber ids:
~~~

I wrote a parser for the ASCII file that automatically creates the test method library hash required by the V93K test interface:

~~~ruby
tml_hash = import_v93k_tml(@client_version, tml: /ccf\.*93k/)
add_tml :amd93k, tml_hash
~~~

It allows users to filter out some desired test method using regular expressions.  Do you think this would be a good PR for origen_testers?

thx

info-rchitect:
@rlaj @ginty @redxeth @coreyeng

ginty:
Yep, I think that would be a good addition

PR #152, opened 2020-04-04
ac_tml and dc_tml conflict solution

info-rchitect:
Hi,

The issue is due to origen_testers having native dc_tml and ac_tml test method libraries that are named the same as the Advantest libraries.  The origen_testers test methods do not match the Advantest test method parameters, preventing users from programming certain test method parameters.  For example in DcTml.Continuity:

origen_testers:

~~~ruby
            continuity:        {
              pinlist:               [:string, '@'],
              test_current:          [:current, 10.uA],
              settling_time:         [:time, 1.ms],
              measurement_mode:      [:string, 'PPMUpar', %w(PPMUpar ProgLoad)],
              polarity:              [:string, 'SPOL', ['SPOL' 'BPOL']],
              precharge_to_zero_vol: [:string, 'ON', %w(ON OFF)],
              test_name:             [:string, 'passVolt_mv'],
              output:                [:string, 'None', %w(None ReportUI ShowFailOnly)]
            },
~~~

Advantest (parsed from ASCII file created by tmm_console utility):

~~~ruby
 "result0"=>[:double, ""],
 "result1"=>[:double, ""],
 "result2"=>[:double, ""],
 "result3"=>[:double, ""],
 "pinlist"=>[:string, ""],
 "testCurrent"=>[:string, ""],
 "settlingTime"=>[:string, ""],
 "measurementMode"=>[:string, ""],
 "polarity"=>[:string, ""],
 "prechargeToZeroVol"=>[:string, ""],
 "testName"=>[:string, ""],
 "output"=>[:string, ""]}
~~~

### DIFF
--- a/lib/origen_testers/basic_test_setups.rb
+++ b/lib/origen_testers/basic_test_setups.rb
@@ -81,7 +81,7 @@ module OrigenTesters
         line = flow.test(ins, options)
         { test_instance: ins, flow_line: line, patset: pset }
       elsif tester.v93k?
-        tm = test_methods.ac_tml.ac_test.functional_test
+        tm = test_methods.ac_tml_native.ac_test.functional_test
         ts = test_suites.run(name, options)
         ts.test_method = tm
         ts.pattern = pattern
--- a/lib/origen_testers/smartest_based_tester/base/test_methods.rb
+++ b/lib/origen_testers/smartest_based_tester/base/test_methods.rb
@@ -7,8 +7,8 @@ module OrigenTesters
 
         require 'origen_testers/smartest_based_tester/base/test_methods/base_tml'
         require 'origen_testers/smartest_based_tester/base/test_methods/limits'
-        autoload :AcTml, 'origen_testers/smartest_based_tester/base/test_methods/ac_tml'
-        autoload :DcTml, 'origen_testers/smartest_based_tester/base/test_methods/dc_tml'
+        autoload :AcTmlNative, 'origen_testers/smartest_based_tester/base/test_methods/ac_tml'
+        autoload :DcTmlNative, 'origen_testers/smartest_based_tester/base/test_methods/dc_tml'
         autoload :CustomTml, 'origen_testers/smartest_based_tester/base/test_methods/custom_tml'
 
         attr_accessor :flow, :collection
@@ -34,13 +34,13 @@ module OrigenTesters
         end
 
         # Returns the AC test method library
-        def ac_tml
-          @ac_tml ||= AcTml.new(self)
+        def ac_tml_native
+          @ac_tml_native ||= AcTmlNative.new(self)
         end
 
         # Returns the DC test method library
-        def dc_tml
-          @dc_tml ||= DcTml.new(self)
+        def dc_tml_native
+          @dc_tml_native ||= DcTmlNative.new(self)
         end
 
         # Creates an accessor for custom test method libraries the first time they are called
--- a/lib/origen_testers/smartest_based_tester/base/test_methods/ac_tml.rb
+++ b/lib/origen_testers/smartest_based_tester/base/test_methods/ac_tml.rb
@@ -2,7 +2,7 @@ module OrigenTesters
   module SmartestBasedTester
     class Base
       class TestMethods
-        class AcTml < BaseTml
+        class AcTmlNative < BaseTml
           TEST_METHODS = {
             frequency_by_digital_capture: {
               class_name:           'Frequency_byDigitalCapture',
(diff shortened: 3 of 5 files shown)


ginty:
>  origen_testers having native dc_tml and ac_tml test method libraries that are named the same as the Advantest libraries

Well, these are Origen mappings of the ADV libraries, with the parameters and their defaults lifted from the SMT documentation.

What's the tmm_console utility?

info-rchitect:
> origen_testers having native dc_tml and ac_tml test method libraries that are named the same as the Advantest libraries
> 
> Well, these are Origen mappings of the ADV libraries, with the parameters and their defaults lifted from the SMT documentation.
> What's the tmm_console utility?

https://github.com/Origen-SDK/origen_testers/issues/70

ginty:
Well, I'm not seeing anything here to convince me that Origen is wrong and should be changed.
This console utility output should not override the actual documentation of these libs. It looks like this tool does not report defaults or contraints on parameter values, and it is including some result fields which are not part of the test method's API.
The solution should be that you don't parse these libs because Origen already provides them.
If you want to enhance by adding aliases to the existing implementation so that camel cased references can be used that would be fine.

info-rchitect:
> Well, I'm not seeing anything here to convince me that Origen is wrong and should be changed.
> This console utility output should not override the actual documentation of these libs. It looks like this tool does not report defaults or contraints on parameter values, and it is including some result fields which are not part of the test method's API.
> The solution should be that you don't parse these libs because Origen already provides them.
> If you want to enhance by adding aliases to the existing implementation so that camel cased references can be used that would be fine.

Parsing this output was a solution proposed by Advantest applications.  Whatever documentation Advantest has cannot be more accurate then the output of the tmm_console utility.  I would like to compare the documentation versus this ASCII file.  Can you paste in a snippet of the documentation for this Contnuity test method?  The camel case is because our users pushed back very hard on using Ruby test method parameter names.  However it isn't really the issue here.

priyavadan:
@ginty @info-rchitect the reality of the question in my opinion is, what is more accurate? TDC for SmartTest or the output of the tmm_console utility?

If I recall correctly, TDC has only the following parameters listed for the DcTml.Continuity library

- pinlist
- testCurrent
- settlingTime
- measurementMode
- polarity
- prechargeToZeroVol
- testName
- output

priyavadan:
to @info-rchitect 's point though, the TDC also states that tmm_console utility dumps the detail information of all the available test methods for this device.

info-rchitect:
Would something like this work?

~~~ruby
        def dc_tml(options)
          options = {
            force_custom_tml: false
          }.merge(options)
          if options[:force_custom_tml]
            @dc_tml_native ||= DcTmlNative.new(self)
          else
            method_missing :dc_tml
          end
        end
~~~

The idea being the user interface could pass the `force_custom_tml` option to have origen_testers ignore the default TML.

rlaj:
Based on Continuity.cpp in the dc_tml, the results[0:4] are valid parameter options so we should just add them to Origen.

priyavadan:
@info-rchitect can we update this PR and include results0..3 in the default parameters for this TM? With that, I think we can ignore parsing this test method out of the output of the tmm_console.
thoughts?

rlaj:
@priyavadan There should not be a camel case issue, what is the specific problem?

priyavadan:
@rlaj asaik, the ASCII file we parse shows a Continuity test method (capitalized) that has a list of test method params --> the tmm_console utility output.

info-rchitect:
Don't we need a full audit of every test method in dc_tml and ac_tml?

ginty:
The bottom line for me is that we are talking about a standard library that Origen is trying to provide support for out of the box.
If you know that Origen's implementation is wrong, then please submit a PR with the corrections.

It is not in the spirit of this open source effort to instead add a hook to allow you to bypass it so that you can use your own (better) internal version of the same thing while leaving the rest of the community with an inferior version.

Similarly, if you really want to be able to type `test_methods.dc_tml.dc_test.generalPmu` instead of `test_methods.dc_tml.dc_test.general_pmu`, then please submit a PR with the alias additions to support that.

rlaj:
Sorry, I guess I'm still not following issue here.  I thought we were just talking about missing parameters.  I've previously used pretty much all of the Advanest canned methods that are supplied by Origen with no issues.  Way back there was issues with camel case, but that was resolved and afaik the parameters (as output by Origen) now match to what SmarTest would create natively.

ginty:
@rlaj, that is pretty much the issue (that some parameters may be missing).
Additionally, I believe there is complaint that users want to reference things like `generalPmu` in their **Origen application code** (i.e. use the same case as in C++) , instead of `general_pmu`.
I don't think there is suggestion that the code produced by Origen has the wrong case.

info-rchitect:
https://github.com/Origen-SDK/origen_testers/issues/153 is the new implementation

PR #154, opened 2020-04-09
enable passing in Camelcase test methods and test method parameters

info-rchitect:


info-rchitect:
I was able to use test method parameter aliases to program a camel-case Continuity test method using non-snakecase test method parameters.

~~~ruby
      # User parameters file
      jtag.define_params :bug_check do |p|
        p.tm.pinlist = 'TDO'
        p.tm.testCurrent = '60[uA]'
      end
~~~

~~~ruby
# User flow file
continuity :bug_check, tml: :dc_tml
~~~

Here is the generated content in the V93K flow file:

~~~ascii
tm_jtag_bug_check:
  "measurementMode" = "PPMUpar";
  "output" = "None";
  "pinlist" = "TDO";
  "polarity" = "SPOL";
  "prechargeToZeroVol" = "ON";
  "settlingTime" = "0.001[s]";
  "testCurrent" = "1.0e-05[A]";
  "testName" = "passVolt_mv";
~~~

This PR also fixed a bug in the continuity test method in the dc_tml library where a parameter default did not properly define an array of default strings.

lib/origen_testers/smartest_based_tester/base/test_methods/base_tml.rb
```diff
@@ -12,6 +12,11 @@ def initialize(test_methods)
           end
 
           def method_missing(method, *args, &block)
+            tml_class_name = self.class.name.to_s.split('::')[-1]
+            # This enables passing camel case test methods for dc_tml and ac_tml
+            unless tml_class_name == 'CustomTml'
```
ginty, line 17: Don't understand what this line is doing, why is a specific library name handled here?

lib/origen_testers/smartest_based_tester/base/test_methods/base_tml.rb
```diff
@@ -12,6 +12,11 @@ def initialize(test_methods)
...
+              method = method.to_s.underscore.to_sym if camelcase?(method)
```
ginty, line 18: I don't think this check for camelcase should be here.
I think this would be better written as something along the lines of - "if the test method as given does match a definition, the lowercase and underscore it and try again".
One of the concerns I have is corner cases where I have seen test method names be inconsistent (I think even in ADV code), between using camelCase and PascalCase.
Taking a less camel case specific approach per above should handle that.

lib/origen_testers/smartest_based_tester/base/test_methods/base_tml.rb
info-rchitect: @ginty The logic prevents altering any custom TMLs.  My understanding is that a custom TML should be handled as-is.

lib/origen_testers/smartest_based_tester/base/test_methods/base_tml.rb
info-rchitect: @ginty OK let me rework it.

info-rchitect: @ginty Is this what you envisioned?

~~~ruby
          def method_missing(method, *args, &block)
            if definitions[method]
              instantiate_test_method(method, args)
            else
              method = method.to_s.underscore.to_sym
              if definitions[method]
                instantiate_test_method(method, args)
              else
                super
              end
            end
          end
~~~

lib/origen_testers/smartest_based_tester/base/test_methods/base_tml.rb
ginty: I see, sorry I missed the line before, I didn't realize this was a Ruby class reference.
Anyhow, I think we should treat custom TMLs the same.

lib/origen_testers/smartest_based_tester/base/test_methods/base_tml.rb
ginty: Yep, exactly like that

info-rchitect:
Hi,

I have setup the test method parameter aliases based on this documentation:

ginty:
> Has anyone used the test method aliases? Are there any examples in origen test?

Yep:

https://github.com/Origen-SDK/origen_testers/blob/89fe28de4a676144bd30a527755531b77407a2b8/lib/origen_testers/test/custom_test_interface.rb#L65

https://github.com/Origen-SDK/origen_testers/blob/89fe28de4a676144bd30a527755531b77407a2b8/lib/origen_testers/test/custom_test_interface.rb#L22

https://github.com/Origen-SDK/origen_testers/blob/89fe28de4a676144bd30a527755531b77407a2b8/approved/v93k/testflow/mfh.testflow.group/custom_tests.tf#L11

info-rchitect:
> Has anyone used the test method aliases? Are there any examples in origen test?
> 
> Yep:
> 
>   
>     
>       origen_testers/lib/origen_testers/test/custom_test_interface.rb
>     
>     
>          Line 65
>       in
>       89fe28d
>     
>     
>     
>     
> 
>         
>           
>            my_arg2_alias: :my_arg2 
>         
>     
>   
> 
> 
> 
>   
>     
>       origen_testers/lib/origen_testers/test/custom_test_interface.rb
>     
>     
>          Line 22
>       in
>       89fe28d
>     
>     
>     
>     
> 
>         
>           
>            ti.my_arg2_alias = 'CURR' 
>         
>     
>   
> 
> 
> 
>   
>     
>       origen_testers/approved/v93k/testflow/mfh.testflow.group/custom_tests.tf
>     
>     
>          Line 11
>       in
>       89fe28d
>     
>     
>     
>     
> 
>         
>           
>            "myArg2" = "CURR";

Thanks @ginty.  So from my audit, it seems like I am doing it correctly.  Do you agree?

ginty:
I think so, it could be a bug that aliases are not currently considered by respond_to.

info-rchitect:
> I think so, it could be a bug that aliases are not currently considered by respond_to.



> I think so, it could be a bug that aliases are not currently considered by respond_to.

@ginty, I think the issue is that the aliases defined in my test interface are not getting passed in for some reason.  Would you expect the code below to return a hash versus `nil`?

image

I was able to workaround the issue with this code in the test method class.

~~~ruby
          options[:attrs].each do |k, v|
            accessor = "#{k}="
            if respond_to?(accessor)
              send(accessor, v)
            else
              accessor = "#{k.to_s.underscore}="
              send(accessor, v) if respond_to?(accessor)
            end
          end
~~~

info-rchitect:
> This looks good.
> Sorry, I took a while to chew it over because I was thrown by the fact that you have a "if doesn't respond to" branch to handle the camel cased name.
> I was confused because I couldn't work out why it wouldn't respond to the camel cased method defined in an alias, then I realized that the whole point of this new update is so that differently cased variants don't have to be declared as aliases. Doh!

Thx @ginty.  Yes I hope this enhancement actually reduces the amount of manual aliasing needed in the end user test interface.

PR #156, opened 2020-05-01
Allow complete control of generator's output path

CodyBriscoe:
Our interface currently moves files after generation to work around the auto inserted subdirectory into the output path. 

This PR allows a user's interface to disable that functionality, by initializing a flag in the default interface to true and then setting it in the user's interface.


CodyBriscoe:
I'll work on adding a case for this to spec/example flow to maintain coverage if we decide to bring this functionality in.

priyavadan:
@CodyBriscoe go ahead and add the spec tests to this. 

The implementation looks good to me. We could probably think of enhancing it in a future PR. @ginty thoughts?

ginty:
Thanks for the PR!

I'm wondering if it would be more useful to make the sub-directory editable rather than just an on/off switch?

There seems to be an instance variable called `@subdirectory` on the flow which contains the offending `'testflow/mfh.testflow.group'`.
Would it not be better to add an `attr_accessor` for that and then you could do `flow.subdirectory = ""` if you wanted no subdir?

spec/v93k_dir_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'spec_helper'
+
+describe 'V93K unique test name generation' do
```
priyavadan, line 3: @CodyBriscoe this needs to be renamed to reflect that this is a v93k output directory spec test

priyavadan:
I believe the intention is to have the capability defined in the interface to specify the sub directory the output files should be placed in. 

If not specified, the default directory structure is used to ensure backwards compatibility.  In our use case, we are able to set the interface.flow_subdirectory to " " in our interface. 

The spec tests that were added covers both the use cases from above.

priyavadan:
discussed this with @CodyBriscoe and we may now have a much simpler way that @ginty refers to in his comment above.

priyavadan:
@ginty thanks

@CodyBriscoe I will merge and make the testers release shortly

PR #157, opened 2020-05-06
added support for v93k for loop

info-rchitect:
Open questions:

1. ATP loop method accepts start, stop, increment, and optional flag.  If the user supplies the optional flag should the stop argument be ignored?
2. If the optional flag argument is supplied, should the flag be set to the stop value on the line before the for loop?

thx


program/flow_control.rb
```diff
@@ -521,4 +521,13 @@
     func :read1, id: :ta5, bin: 10, number: 60000
     func :erase1, if_all_sites_passed: :ta5, bin: 12, number: 60010
   end
+
+  if tester.v93k?
+    loop from: 1, to: 5, step: 1, var: "loop" do
+      func :loop_check, test_number: 60099
+    end
+    loop from: 1, to: 5, step: 1 do
+      func :loop_check_no_var, test_number: 60099
```
priyavadan, line 530: @info-rchitect  this will fail generation, test number needs to be unique here

info-rchitect:
Cancelling this, pls go [here ](https://github.com/Origen-SDK/origen_testers/pull/158)for a cleaner PR

PR #158, opened 2020-05-06
new v93k loop PR

info-rchitect:
Open questions:

- ATP loop method accepts start, stop, increment, and optional flag. If the user supplies the optional flag should the stop argument be ignored?
- If the optional flag argument is supplied, should the flag be set to the stop value on the line before the for loop?

thx


approved/v93k/testflow/mfh.testflow.group/flow_control.tf
```diff
@@ -3764,6 +3794,16 @@ test_flow
     else
     {
     }
+    for @index = 0; @index < @loop ; @index = @index + 1; do
+    test_number_loop_increment = 0
```
ginty, line 3798: What's this test_number_loop_increment ?

approved/v93k_smt8/OrigenTesters/flows/FLOW_CONTROL.flow
```diff
@@ -1751,5 +1765,15 @@ flow FLOW_CONTROL {
         else
         {
         }
+        for @index = 0; @index < @loop ; @index = @index + 1; do
```
ginty, line 1768: I doubt this is valid smt8, its fine to not support it on SMT8 yet, but in that case either output nothing and emit a warning, or else raise an error rather than generating garbage.

priyavadan: @info-rchitect  can you look at this? adding it in a tester.smt7? check should be okay

approved/v93k/testflow/mfh.testflow.group/flow_control.tf
priyavadan: @info-rchitect is this something that just needs to be removed from here?

info-rchitect:
@rlaj Can you please review this PR as it uses the ATP `loop` method you created.

lib/origen_testers/smartest_based_tester/base/flow.rb
```diff
@@ -264,6 +264,25 @@ def capture_lines
...
 
+        def on_loop(node)
+          loop_start = node.children[0]
+          loop_end = node.children[1]
+          loop_inc = node.children[2]
+          index_flag = node.children[3] if node.children[3].is_a?(String)
+          if index_flag.is_a?(String)
+            line "for @#{index_flag} = #{loop_start}; @#{index_flag} < #{loop_end} ; @#{index_flag} = @#{index_flag} + #{loop_inc}; do"
+          else
+            line "for @index = #{loop_start}; @index < #{loop_end} ; @index = @index + #{loop_inc}; do"
```
ginty, line 275: I think you need to generate something more unique that just `@index`, e.g. consider the case of nested loops.
If you look around in here there should be other examples of generating unique ids.

info-rchitect: @ginty Is that the responsibility of Origen?  Seems like that should be up to the user similar to how one would use different variables in a C for loop.

ginty: Yes, this is the kind of detail that a framework should take care of

ginty: Also consider the Origen API is trying to be as ATE-agnostic as possible. If a user only wants to loop X times, then they should be able to express that without worrying about flow vars. Such things may not even make sense on other platforms.

info-rchitect: > Also consider the Origen API is trying to be as ATE-agnostic as possible. If a user only wants to loop X times, then they should be able to express that without worrying about flow vars. Such things may not even make sense on other platforms.

The current ATP loop API allows the user to pass an integer or a string representing a flow variable for the `to` option.  Wouldn't we just handle that in each of the tester generators?  If a particular generator does not use an option available in the ATP flow method then it errors and instructs the user.

ginty: This is the way Origen works man, with the current loop API I would and should expect this to just work:

~~~ruby
loop from: 1, to: dut.number_of_x, step: 1 do
    loop from: 1, to: dut.number_of_y, step: 1 do
      func :some_test
    end
end
~~~

If the v93k needs a flow var to be defined to enable that to work, then its Origen's responsibility to create one (if the user doesn't care about what it's called), and make sure what it is creating will work in cases like the above.

PR #160, opened 2020-07-10
added program comment char to SMT testers

info-rchitect:
This PR should force the flow file diff to ignore the proper comment characters for SMT7 and SMT8.  Not sure if I can test this fully as the diff code is in the origen core.  Any ideas on testing this?

### DIFF
--- a/lib/origen_testers/smartest_based_tester/base.rb
+++ b/lib/origen_testers/smartest_based_tester/base.rb
@@ -70,6 +70,9 @@ module OrigenTesters
       # format (SMT8 only)
       attr_accessor :zip_patterns
 
+      # Program comment character that affects what portions of a gnerated flow file to ignore during diff
+      attr_reader :program_comment_char
+
       def initialize(options = {})
         options = {
           # whether to use multiport bursts or not, if so this indicates the name of the port to use
@@ -99,8 +102,10 @@ module OrigenTesters
         @min_repeat_loop = 33
         if smt8?
           @pat_extension = 'pat'
+          @program_comment_char = '//'
         else
           @pat_extension = 'avc'
+          @program_comment_char = '--'
         end
         @compress = true
         # @support_repeat_previous = true
--- a/lib/origen_testers/origen_ext/generator/flow.rb
+++ b/lib/origen_testers/origen_ext/generator/flow.rb
@@ -183,6 +183,85 @@ module Origen
         [flow_comments, comments]
       end
 
+      def header
+        Origen.tester.pre_header if Origen.tester.doc?
+        inject_separator
+        if $desc
+          c2 'DESCRIPTION:'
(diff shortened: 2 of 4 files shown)


priyavadan:
@info-rchitect looks good. 

The code that compiles the test flow files, uses the [Origen.app.program_comment_char](https://github.com/Origen-SDK/origen_testers/blob/282e99230d408277b0358fc952aa983205e35b0b/lib/origen_testers/generator.rb#L213)

to verify this, you need to, once this is merged and a newer release is available, update Origen to use that version and add a spec test in there.

lib/origen_testers/smartest_based_tester/base.rb
```diff
@@ -70,6 +70,9 @@ class Base
       # format (SMT8 only)
       attr_accessor :zip_patterns
 
+      # Program comment character that affects what portions of a gnerated flow file to ignore during diff
+      attr_reader :program_comment_char
+
```
ginty, line 75: Where does this actually get used?

priyavadan: @ginty [here](https://github.com/Origen-SDK/origen_testers/blob/282e99230d408277b0358fc952aa983205e35b0b/lib/origen_testers/generator.rb#L213), I think

redxeth:
@info-rchitect was this ever merged?  I encountered this issue today and was going about updating it in another PR I'm working on when I discovered this PR.

I'm doing it a different way, adding:

    @program_comment_char = ['print_dl', '//'] 
    
to `lib/origen_testers/smartest_based_tester/base.rb` which corresponds to where it is for the IGXL-based testers.

Also i want to add assign methods for both `comment_char` and `program_comment_char` the API so that any user can change these if desired for special situations.

Should I pick up this PR -- would like to get this merged ASAP.

Thoughts? @ginty @priyavadan

redxeth:
I can also add these changes to my https://github.com/Origen-SDK/origen_testers/pull/178 if we don't wantto revive this PR.

redxeth:
Ugh-- finding more problems besides just a simple change related to the decompiler and how it uses the `comment_char`.

Need to get [PR178](https://github.com/Origen-SDK/origen_testers/pull/178) merged sooner rather than later so will prob keep them separate.

redxeth:
Created [PR180](https://github.com/Origen-SDK/origen_testers/pull/180) to support update for `program_comment_char`.  Since I don't really understand the decompiler use case code, will leave it for someone who does.  This PR include exposing `program_comment_char` and `comment_char` in case user wants to redefine it from app level.

redxeth:
@info-rchitect this has been fixed now on [PR180](https://github.com/Origen-SDK/origen_testers/pull/180).  So perhaps you can close/delete this PR?

PR #162, opened 2020-07-21
Allow the SMT8 limits 'test_text' value to be specified

ginty:
Allows the value in the 'test_text' column in the SMT8 limits file to be specified from the flow/interface.


lib/origen_testers/smartest_based_tester/base/limits_file.rb
```diff
@@ -252,9 +253,12 @@ def line(options)
             # "Test Number"
             l << f(options[:test_number])
             # "Test Text"
-            # l << f(options[:bin_s_name] || options[:bin_h_name])
-            names = ["#{options[:suite_name]}", "#{options[:test_name]}"]
-            l << f(names.uniq.join('.'))
+            if options[:test_text]
+              l << options[:test_text]
+            else
```
ThaoHuynhFsl, line 258: Hi @ginty this needs to change to  l << f(options[:test_text]).  cheers

PR #163, opened 2020-07-26
allow for program generation when no variables were created

CodyBriscoe:
I threw together a quick and simple testcase for some unrelated dev work, but ran into an error. I found that if no variables were created by the flow then the origen p command would error out. 

This PR allows for program generation to complete when no system or user variables are generated for the v93k

### DIFF
--- a/lib/origen_testers/smartest_based_tester/base/variables_file.rb
+++ b/lib/origen_testers/smartest_based_tester/base/variables_file.rb
@@ -35,21 +35,25 @@ module OrigenTesters
 
         # What SMT7 calls a flag
         def flags
-          (variables[:all][:referenced_enables] + variables[:all][:set_enables]).uniq.sort do |x, y|
-            x = x[0] if x.is_a?(Array)
-            y = y[0] if y.is_a?(Array)
-            # Need to use strings for the comparison as some flags can be a string and some a symbol
-            x.to_s <=> y.to_s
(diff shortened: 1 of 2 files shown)


priyavadan:
@CodyBriscoe can't think of a reason why a flow would ever have no declarations or variables , the changes look ok to me though!

CodyBriscoe:
@priyavadan agreed that a test program that doesn't have declarations or variables is rare, and most users would rarely if ever have origen fail out on them for this case. Though if this case can't be handled, then, imo, origen_testers is making the statement that if your program does not add a declaration or variable, then it is an invalid program in every scenario and should never continue to generate output for that case. So now adding variables/declarations is a program requirement. 

Now maybe thats a correct statement, I don't have the industry experience to make a judgement on that.

priyavadan:
@CodyBriscoe with that said what's your flow that throws the error? Like I said earlier the code changes are trivial and look okay.

CodyBriscoe:
@priyavadan simple flow that only had a single test in it, a test_variable test, was using it to step through some code to understand where to interrupt the process of adding a test for charz dev.

info-rchitect:
@CodyBriscoe Should we put a spec test in for this?

CodyBriscoe:
@info-rchitect sure, but first want to make sure the consensus is that this PR is a value add

PR #164, opened 2020-08-10
Charz API

CodyBriscoe:
~~Currently marked as draft PR until I can write the spec tests, but wanted to go ahead and get this out in the meantime to start discussion.~~

This PR allows an applications interface to easily add characterization to existing flow files.

The short of it is that an interface would define charz routines, which are objects that contain characterization specific data such as search range, resolution, etc. Then, right after creating the point test (non charz test), the API will handle everything with regards to combining the options of that point test with charz routines, conditional execution based on enables/flags/test result settings, and placement of that test (ie immediately after the point test, end of the flow, or somewhere else).

As I only have access to the V93K, I can't speak for the validity on other testers, though I tried to make sure nothing in this PR is tester specific.

A quick 4 part example with code snippets (Routine/Profile Setup, Interface Adjustments, Flow Example Usage, Pseudo output:

1. Routine/Profile setup
``` ruby
# routines
add_charz_routine :vmin do |routine|
  routine.name = 'cz_vmin_vdd'
  routine.start = 1.0.V
  routine.stop = 0.5.V
  routine.res = 5.mV
  routine.spec = 'VDD'
end
add_charz_routine :vmax do |routine|
  routine.name = 'cz_vmax_vdd'
  routine.start = 1.0.V
  routine.stop = 1.5.V
  routine.res = 5.mV
  routine.spec = 'VDD'
end
# profile
add_charz_profile :fail_searches do |profile|
  profile.name = 'fail_searches'
  profile.on_result = :on_fail
  profile.routines = [:vmin, :vmax]
  profile.enables = { ['$VmaxEnable'] => [:vmax] }
end

```
2. Adjustments to my local interface to incorporate the API
``` ruby
Module MyApp
  class Interface
    ...
    include OrigenTesters::Charz
    ...
    def shutdown(options = {})
      if options[:top_level]
        generate_eof_charz_tests(options)
      end
    end
    ...
    # my interface's implementation of adding tests to the flow
    def add_to_flow(instance, test_obj, options)
      unless charz_only? and !options[:charz_test] # charz_test is specific to my interface
          set_conditional_charz_id(instance, options) # adds an id to the test if charz is result dependent
        super # parent version actually adds the test
      end
      insert_charz_tests(options.merge(parent_test_name: instance.name)) do |options|
        # method specific to my interface that handles combining a routine and a point test to
        # create the actual test that gets inserted into the flow, ie:
        # copies over the pattern, and other test info, handles what a test should with a routine.start value, etc
        # the current routine gets passed to this block in options[:current_routine]
        # other options are the same options that were previously passed to super to create the point test
        insert_current_charz_test(options)
      end
    end
```
3. Usage in the flow files
``` ruby
# :fail_searches is the profile name, but I could have passed a routine(s) instead
# updating the placement to override what was in the profile (which was nothing, it defaults to :inline )
charz_on :fail_searches, { placement: :eof }
crest :func, pat: :my_burst1, test_point: :vnom
charz_off
crest :func, pat: :my_burst2, test_point: :vnom
```
4. V93K "Pseudo" Output Snippet (Some edits for brevity, mainly run_and_branch -> run syntax)
```
# this is the end of the flow, the last thing executed was the my_burst2 test
    {
       {
          if @FUNC_CPU_MY_BURST1_FAIL_SEARCHES_9777D19_FAILED == 1 then
          {
             run_and_branch(func_cpu_my_burst1_vnom_cz_vmin_vdd)
             if @VmaxEnable == 1 then
             {
                run_and_branch(32_29_func_cpu_my_burst_123_vnom_cz_vmax_vdd)
             }
          }
       }, open,"func_cpu_my_burst1_vnom charz fail_searches", ""
    }, open,"End of Flow Charz Tests", ""
```

~~I plan on writing the spec tests/examples in the next few days, once I'm happy with them I'll move this out of the draft state.~~


priyavadan:
I removed the ruby 2.4.7 check as it was having issues with simplecov, similar to how ruby 2.3 has had in the past. 

While installing Ruby 2.4.7, the message says that it is end of life, so I think this should be acceptable.

lib/origen_testers/test/interface.rb
```diff
@@ -86,6 +134,82 @@ def func(name, options = {})
...
+            ts.test_method = tm
+            if tester.smt8?
+              fail 'SMT8 Not Yet Implemented'
+              # ts.spec = options.delete(:pin_levels) if options[:pin_levels]
+              # ts.spec ||= 'specs.Nominal'
+            else
+              ts.levels = options.delete(:pin_levels) if options[:pin_levels]
+            end
+            ts.pattern = 'charz_example'
+            # if block
```
priyavadan, line 183: @CodyBriscoe does this need a clean up?

ginty (changes requested):
Hi @CodyBriscoe,

Apologies for not reviewing this sooner, it looks really nice and is a very welcome addition to Origen.

Could you add some documentation for it? That doesn't need to be much more than just copying what you already have in the PR description.
I would recommend adding it to the V93K page for now and we can thinking about moving it in future if and when we support other platforms - https://origen-sdk.org/origen/guides/program/v93k/
The source for that is in origen_testers - https://github.com/Origen-SDK/origen_testers/blob/master/templates/origen_guides/program/v93k.md.erb

You can review the output locally by running `origen web compile templates/origen_guides`.

With regards to the API, is the manual call to `generate_eof_charz_tests(options)` required? Is there any case where you would not want to call that? If not it would be nicer to automate it.

Also would it be possible to give more of an example of what `insert_current_charz_test` does in the documentation?
From reading this currently I am not clear on whether this picks up all the heavy lifting to generate cz tests, or if I still have a lot of work to do within my own version of that method to enable it.

Thanks a lot for the great work so far!

ginty:
Just one other thought, in addition to this:

~~~ruby
charz_on :fail_searches, { placement: :eof }
crest :func, pat: :my_burst1, test_point: :vnom
charz_off
~~~

Could we also support this?


~~~ruby
charz_on :fail_searches, { placement: :eof } do
  crest :func, pat: :my_burst1, test_point: :vnom
end
~~~

That's more aligned with how the other flow APIs work, like `if_enable`, `if_failed`, etc.

CodyBriscoe:
Thanks for the feedback @ginty and sure thing I'll have some documentation up for review soon.

I haven't made the generate_eof_charz_tests automatic, as I was trying to ensure the end user had as much control as possible. Currently scoping out where to add the call to that method so that nothing breaks if a user doesn't want to include OrigenTesters::Charz in their interface, and still has access to check if they want to customize the grouping aspect or naming. Will get back to you on this.

As far the `charz_on do` I'm a fan of this, and it makes sense so I'll add that capability. However we should still keep the non block syntax available, or otherwise there are situations where you'd have to repeat tests, for example:
``` ruby
if add_charz?
  charz_on :my_profile do
    production_test1
    production_test2
  end
else
  production_test1
  production_test2
end
```
where with the non block syntax:
``` ruby
charz_on :my_profile if add_charz?
production_test1
production_test2
charz_off if add_charz?
```

CodyBriscoe:
hey @ginty, added block syntax to enable `charz_on :profile do/end`, eof charz test generation is now automatically called right before shutdown, and a characterization guide is written and added to origen guides. 

I ended up making the charz guide its own page, as its not really V93K specific, its just the only environment I can prove it out on myself, but it should work for all testers. I did make a note on the guide that its only been tested on v93k smt7.

If these updates check out to y'all its ready to merge

CodyBriscoe:
Also added a section on #insert_charz_tests to try and clarify what all that method does for you

ginty (approved):
Thanks for the updates @CodyBriscoe, the charz API guide looks great and definitely helps to make it clearer now how it is not V93K-specific.

PR #167, opened 2020-10-14
Adding IGXL node to permit  moving any device conditions on a flow line to a group condition

redxeth:
@ThaoHuynhFsl has asked me to make a change for one of the project he's worked on before.  

The change is to add an igxl node to permit flows to take advantage of the group conditions instead of just the device conditional.

Not sure if this will fly or not, given that it's a tester-specific thing.  Want to see how the team wants to properly handle it.

--Daniel

### DIFF
--- a/lib/origen_testers/atp/flow.rb
+++ b/lib/origen_testers/atp/flow.rb
@@ -419,6 +419,10 @@ module OrigenTesters::ATP
             end
           end
 
+          if options[:igxl]
+            children << n1(:igxl, options[:igxl])
+          end
+
           if before_on_fail
             on_fail_node = on_fail(before_on_fail)
             if options[:on_fail]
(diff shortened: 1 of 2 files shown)


redxeth:
All, in implementing a test case for this update, I found that the capability to put something in the group condition columns for IGXL is already there-- you indicate `if_any_site_failed` instead of `if_failed` on a test.   

Still, do people think that having a tester-specific node like this is something we should have ?

pderouen:
I don't see an issue with having an ig-xl specific node. Do you think this PR is still necessary though?

redxeth:
> I don't see an issue with having an ig-xl specific node. Do you think this PR is still necessary though?

I don't think so-- we were able to fix it via an OrigenTesters update.  Still, am consulting with the relevant team.

Thanks.

redxeth:
Thao said we can kill it.

PR #168, opened 2020-10-22
Igxl defer

pderouen:
Allow defer limits to be used for nodes that have all limits entered as sub tests


priyavadan:
@pderouen should we add an example use case in the igxl specs ?

pderouen:
Not sure why the coverage dropped after ADDing tests. I would have thought it was the code added to the test interface. But, the report shows the coverage on that file went up.

priyavadan (approved):
looks good to me, did not understand either why coverage went down a bit.

pderouen:
Thanks, I will merge this later today if there are no objections.

PR #170, opened 2020-12-03
fixed a corner case bug that causes the generator to hang

pderouen:
In the case where a pattern is created but no cycles are generated, the header method would hang. Not something you'd normally expect to be done. But, in the case where you're taking advantage of the 64 cycle minimum cycle count, you can still get a valid pattern without ever generating a cycle.


info-rchitect (approved):
LGTM

pderouen:
@info-rchitect thanks for the speedy review!!

ginty:
Looks good

PR #172, opened 2021-01-20
Prevent test number duplication (violates STDF standards)

pderouen:
This change was requested to bring an existing app up to STDF standards compliance. It does open the possibility of creating a scenario where the auto-incremented test numbers over run the test number for a subsequent test. Either case the duplicated test numbers are not allowed and cause issues with STDF output.

The other option for correcting the test numbers in an app is to use sub tests to create use-limit lines. In this case the app can define the tnum or leave it blank.


redxeth:
@pderouen  Sorry, it looks like a small change.  I've asked Fazila to take a look since it's her request.

redxeth (approved):
Looks good to me-- Fazila trialed this and confirmed this solves the test number on use-limit issue.

PR #173, opened 2021-04-15
get unfrozen string before modifying

pderouen:
I was unable to create a case to duplicate the error. But, there is a case where the object passed to <code>clean_flag</code> is frozen (immutable). The string created by <code>to_s</code> will also be frozen in that case. The <code>.dup</code> method returns a mutable copy of the string.

### DIFF
--- a/program/flow_control_flag_bug.rb
+++ b/program/flow_control_flag_bug.rb
@@ -1,0 +1,25 @@
+# Flow to exercise the Flow Control API
+#
+# Some of the other flows also cover the flow control API and those tests are used
+# to guarantee that the test ID references work when sub-flows are involved.
+# This flow provides a full checkout of all flow control methods.
+Flow.create interface: 'OrigenTesters::Test::Interface', flow_name: "Flow Control Testing" do
+  flow.flow_description = 'Flow to exercise the Flow Control API' if tester.v93k?
+
+  self.resources_filename = 'flow_control'
+
+  log "Mixed-case manual flags"
+  test :test1, on_fail: { set_flag: :$My_Mixed_Flag }, continue: true, number: 51420
+  test :test2, if_flag: "$My_Mixed_Flag", number: 51430
+  unless_flag "$My_Mixed_Flag" do
+    test :test3, number: 51440
+  end
+  
+  log "Mixed-case manual flags - induce frozen string error"
+  test :test4, on_fail: { set_flag: :$My_Mixed_Flag }, continue: true, number: 51450
+  test :test5, if_flag: "$My_Mixed_Flag", number: 51460
+  unless_flag "$My_Mixed_Flag" do
+    test :test6, number: 51470
+  end
+
+end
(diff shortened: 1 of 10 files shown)


info-rchitect:
@pderouen Do you want to add a unit test to check for this corner case?  Otherwise LGTM.

program/flow_control_flag_bug.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  log "Mixed-case manual flags"
+  test :test1, on_fail: { set_flag: :$My_Mixed_Flag }, continue: true, number: 51420
+  test :test2, if_flag: "$My_Mixed_Flag", number: 51430
+  unless_flag "$My_Mixed_Flag" do
+    test :test3, number: 51440
+  end
+  
+  log "Mixed-case manual flags - induce frozen string error"
+  test :test4, on_fail: { set_flag: :$My_Mixed_Flag }, continue: true, number: 51450
```
pderouen, line 19: Using the same flag name with a second set_flag is what leads to the frozen string bug.

pderouen:
@info-rchitect I'm not sure how to get at the <code>clean_flag</code> method from a spec test. But, I was finally able to duplicate the behavior and added an example for it.

info-rchitect (approved):
LGTM

redxeth:
@pderouen for this bug, what do you expect the proper output should look like? 

for UF today the `flow_control_flag_bug.rb` flow generates  :

~~~text           
Opcode    Parameter      Fail                Sense   Condition     Name 
Test      test1          My_Mixed_Flag_0
defaults                 My_Mixed_Flag
flag-true My_Mixed_Flag                              flag-true     My_Mixed_Flag_0
Test      test2                                      flag-true     My_Mixed_Flag
Test      test3                              not     flag-true     My_Mixed_Flag
Test      test4          My_Mixed_Flag_1
flag-true My_Mixed_Flag                              flag-true     My_Mixed_Flag_1
Test      test5                                      flag-true     My_Mixed_Flag
Test      test6                              not     flag-true     My_Mixed_Flag
~~~

It appears that the on_fail/set_flag is properly incrementing the flag name, but there are some unnecessary lines (defaults, flag-true lines) as well as the wrong flag used for test2/3/5/6 .  **Is that your expectation also?**

So it seems more than just the 'frozen strings' (`My_Mixed_Flag` not updating to include `_0` and `_1` but it seems other stuff is extraneous.  

Also, looking at it now I'm not entirely sure that `if_ran`/`unless_ran` also works?

@priyavadan @ginty @chrisnappi @info-rchitect

pderouen:
@redxeth that uflex output produces the functionality that was specified in the test. This PR was just to resolve a hard crash. The flag string was intentionally frozen. When the _0 and _1 flag strings were being created by duplicating the frozen string, which produced frozen strings that couldn't have the _0 etc appended. Fixing that was the full scope of this PR. None of the underlying functionality was touched.

You should consider starting a new [discussion](https://github.com/Origen-SDK/origen_testers/discussions/landing).

PR #174, opened 2021-06-28
Add overlay label capability for SmarTest test patterns

redxeth:
All,

@ThaoHuynhFsl implemented a change in a local fork of this repo but I need the change in the master branch / gem.

All it does when a label is requested is to include a comment label that can searched for via C code (SMT7) or Java (SMT8) GetComment capability.  This way you can obtain the vector offset for the overlay.

Thanks,

Daniel


priyavadan (approved):
Looks good to me.

I believe the diff based regression testing (approved folder) covers for this use case.

priyavadan:
@redxeth while at it, can you enable  github actions regressions for origen testers? 

Origen example here: https://github.com/Origen-SDK/origen/blob/master/.github/workflows/regression_tests.yml

redxeth:
@priyavadan  sure will do.

redxeth:
All, 

Modified github actions regressions.yml to only cover ubuntu (not windows) and only for Ruby 2.5 and 2.6.  Was seeing failure for ruby 2.7 and windows for the `origen test -c`, the `examples` part.  I do not use Ruby 2.7 nor windows for origen, and the previous travis.yml only covered for ruby 2.5.6 and 2.6.4 (for linux only I assume) so will rever this file likewise.

If someone wants additional coverage for windows or later versions of ruby then that should be a separate ticket/PR i think to enhance the test coverage.

redxeth:
Any thoughts?  Do I need to add github actions ability for windows or not?  I'd have to install windows capability which haven't tried yet.

pderouen:
> 
> 
> Any thoughts? Do I need to add github actions ability for windows or not? I'd have to install windows capability which haven't tried yet.

Yes! Otherwise we won't catch regressions that are only related to Windows. Not a big deal over a short period of time. But, I wouldn't want to drop Windows CI unless we're dropping Windows support.

redxeth:
Yeah-- but we didn't have that before with Travis CI, right?  I can add it but it will delay this PR more since I don't have a windows setup.

pderouen:
I'm fine merging this PR without windows CI running.

pderouen:
I created https://github.com/Origen-SDK/origen_testers/issues/175 to track the Windows issue. I downloaded the main branch and ran the tests from Windows and see several failures.

pderouen (approved):

redxeth:
Awesome thanks @pderouen !   Let me see about installing windows to resolve the CI issues.  In the meantime if you can zip up the output dir you're getting then at minimum I can fix the examples failures.

Issue #175, opened 2021-07-01
Tests fail in Windows


pderouen:
At least 1 example and several spec tests fail when run from Windows

pderouen:
The examples failure in windows is a difference in 93k program generator output. cc comments and '#' comments immediately before a test are not being added as comments to the .tf output.

The issue is in the code that examines the call stack to find the file and line number that produced the flow entry. [This line](https://github.com/Origen-SDK/origen_testers/blob/425aa32f66ebb024427b615495dc2e0ac99984e4/lib/origen_testers/flow.rb#L207) is not finding the correct entry in the call stack when running on Windows.

PR #176, opened 2021-07-02
Windows Updates

pderouen:
Fixes https://github.com/Origen-SDK/origen_testers/issues/175 and a handful of Windows related bugs.


spec/spec_helper.rb
```diff
@@ -1,22 +1,23 @@
...
+#    `where origen`.split("\n").find do |match|
+#      match =~ /(.*)\\bin\\origen$/
+#    end
+#    origen_top = $1.gsub("\\", "/")
+#  else
+#    origen_top = `which origen`.strip.sub("/bin/origen", "")
+#  end
+#
+#  $LOAD_PATH.unshift "#{origen_top}/lib"
+# end
```
redxeth, line 20: So this works on windows now? If so maybe clear out the old comments?

pderouen: That code was checking to see if ../Gemfile exists. If it doesn't exist, it was checking the OS and which origen. None of that code was working correctly from Windows.

It all seems like it became outdated long ago (I don't know of a version of origen_testers that doesn't have Gemfile 1 directory up from spec/)

pderouen:
@redxeth I removed those commented out lines of code from spec_helper

redxeth:
What about the my other comment-- should more of pattern_compilers spec be under running_windows? check? Also I noticed a small indentation issue on https://github.com/Origen-SDK/origen_testers/blob/windows/spec/pattern_compilers_spec.rb#L608.   I guess lint doesn't run on the specs?

pderouen:
> What about the my other comment-- should more of pattern_compilers spec be under running_windows? check? 

I only blocked the failing tests from running on Windows. For those origen_testers is generating an error if called while running on Windows.

> Also I noticed a small indentation issue on https://github.com/Origen-SDK/origen_testers/blob/windows/spec/pattern_compilers_spec.rb#L608. I guess lint doesn't run on the specs?

No, it doesn't run on specs. I noticed and fixed a few indentation issues when I was working on this PR. I'll see what I can do for it.

pderouen:
@redxeth running lint on everything in the specs directory caused rspec to stop working. So, I ran lint just on the 2 files I touched.

redxeth (approved):

spec/pattern_compilers_spec.rb
```diff
@@ -550,60 +550,62 @@ def configs
...
-        dut.v93k_compiler_options[:clean] = true
-        dut.add_pattern_compiler(:id1, :v93k, dut.v93k_compiler_options)
-        dut.pattern_compilers[:id1].run
-      end
-
-      it 'runs existing aiv file' do
-        dut.pattern_compilers.should == {}
-        dut.v93k_compiler_options[:clean] = true
-        dut.add_pattern_compiler(:id1, :v93k, dut.v93k_compiler_options)
+      unless Origen.running_on_windows?
```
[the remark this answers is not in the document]
pderouen: I'm sort of confused as to why pattern compilation is designed to error out when running on Windows. It makes sense for 93k I suppose. But, the compilers for Teradyne patterns are native to Windows. At some point we might want to change the code to check if the compile command is available on the system. Though I don't think there's a way to have them installed on the CI Windows machine. That can be an update for a future PR (or not).

PR #178, opened 2021-08-11
Flow loops v93k

redxeth:
Added simple flow loop control for V93K tester (SMT7 only):

    loop from: 0, to: 5, step: 1, var: '$LOOP_VARIABLE' do
      func :test_myloop, number: 56000
    end

and can do nested:

    loop from: 0, to: 9, step: 2, var: '$LOOP_VARIABLE1'do
      loop from: 1, to: 10, step: 1, var: '$LOOP_VARIABLE2' do
        loop from: 1, to: 5, step: 1, var: '$LOOP_VARIABLE3' do
          func :test_myloop5, number: 56400
        end
      end
    end


redxeth:
For SMT8 I spoke with @ThaoHuynhFSL and he spoke with Advantest indicating that although you CAN do a loop in SMT8:

    for (i: 1..100) {
        test.loop = i;
        test.execute();
    }

It's not recommended due to large overhead associated with each `execute()` run.  It is better to do looping instead in the test method code.

For SMT7 it is preferred to do in the flow to minimize the use of sequencer labels and increasing the limits table dramatically.

redxeth:
Anyway decided to go ahead and do program comment char thing in a separate PR.  Need to get these flow loops merged.

pderouen (approved):
Looks good to me

priyavadan (approved):
LGTM!

redxeth:
Thanks guys!  I just need to make a small update to the Flow API docs--

Question:  since these `origen_guides` get brought over to the main Origen website-- how can you do a local compile to check them?

PR #180, opened 2021-09-10
ADD program comment chars for SMT7/8

redxeth:
C or Java code program comment char not set properly for SMT7/8 test programs.  Since was not set would default to `#` which would make having built-in Origen diffs not be able to skip those comments in that code.

Have added `program_comment_char` for Smartest_based testers.

Also added ability for user to modify the `comment_char`or `program_comment_char` for any tester -- in case a strange case comes up in the future.

### DIFF
--- a/lib/origen_testers/api.rb
+++ b/lib/origen_testers/api.rb
@@ -11,6 +11,7 @@ module OrigenTesters
     attr_accessor :includes
     attr_accessor :comment_level
     attr_accessor :generating
+
     # Get/Set the overlay style
     #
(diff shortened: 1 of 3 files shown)


pderouen:
Seems pretty straightforward to me. Do you plan to add any tests?

redxeth:
Wasn't sure if there were any for the comment chars in general.  Let me check.

redxeth:
Checked that the test case I added does indeed fail if you have the wrong `@program_comment_char` in place.

pderouen (approved):

PR #183, opened 2022-02-24
make the digital_instrument programmable

pderouen:


pderouen:
I hope the regression tests not being able to run any longer with the GA setup isn't going to be a persistent issue.

priyavadan (approved):
Looks good to me, thanks for the updates @pderouen

PR #187, opened 2022-05-18
Feature/flag enable syntax

redxeth:
Added ability to override default syntax behavior where flag or enable word names are made lowercase and any double-scores removed.  

This feature can be used by setting options when instantiating the tester object:

~~~ruby
OrigenTesters::UltraFLEX.new(literal_flags: true, literal_enables: true)

~~~

`literal_flags` option is to force OrigenTesters to not modify the flag names used so that you can do:

~~~ruby
func :read1, id: :Test__Flag1, bin: 10, number: 50000
func :erase1, if_failed: :Test__Flag1, bin: 12, number: 50010
~~~

Should generate flag name `Test__Flag1` in the test program.  NOTE: that this is only evident in IG-XL based testers which use these flags for flow control.  SmarTest based testers use branching natively in the flow without flags so these flag ID names are not evident for them.

Same is true for enable words:
~~~ruby
func :extra_test, if_enable: :Extras__123, number: 50270
~~~

Will generate enable words `Extras__123` for IG-XL based testers.


pderouen (approved):
Looks good to me. Thank you for this update!

PR #190, opened 2023-02-06
Feature/add flag global settings

JamesCreamer:
The initialization block was auto generated for smartest testers. Global flags were needed to persist the value between unit runs for test time reduction. I saw files in the igxl testers block that referred to global specs and I am less familiar with the ultraflex and j750 test setups. Do any blocks of code need to be updated for this use case?


priyavadan:
@JamesCreamer thank you for the updates. Please consider renaming "global" to something more specific, "global_flags" maybe?

Also, please add documentation around the new feature in the guide templates! Thank you again!

priyavadan (approved):
Thank you for the updates @JamesCreamer !

PR #191, opened 2023-04-14
Feature/literal flag options

redxeth:
Added ability to do additonal custom changes to flag names-- in case user has particular requirements.

Flag name modifications: 
- Added ability to substitute the word "FAILED" for any custom word. Also abilty to change "PASSED" and "RAN" as part of flag name also.
- Added ability to put the type of flag (fail/pass/ran) first instead of last in the name.

### DIFF
--- a/approved/j750_literals/flow_control_literals_flow.txt
+++ b/approved/j750_literals/flow_control_literals_flow.txt
@@ -18,26 +18,26 @@ DFF 1.1	Flow Table
 						Test	pgm1	pgm1	50090					Fail									flag-true	p_flw_FLOW_CONTROL_LITERALS_Test__Flag4_06D8931				
 						logprint	Test_that_if_ran_works																					
 						Test	pgm	pgm	50100		10			Fail														
-						defaults										RAN_flw_FLOW_CONTROL_LITERALS_Test__Flag5_06D8931												
-						flag-true	RAN_flw_FLOW_CONTROL_LITERALS_Test__Flag5_06D8931																					
-						Test	read0	read0	50110					Fail									flag-true	RAN_flw_FLOW_CONTROL_LITERALS_Test__Flag5_06D8931				
+						defaults										r_flw_FLOW_CONTROL_LITERALS_Test__Flag5_06D8931												
+						flag-true	r_flw_FLOW_CONTROL_LITERALS_Test__Flag5_06D8931																					
+						Test	read0	read0	50110					Fail									flag-true	r_flw_FLOW_CONTROL_LITERALS_Test__Flag5_06D8931				
 						logprint	Test_the_block_form_of_if_ran																					
 						Test	pgm	pgm	50120		10			Fail														
-						defaults										RAN_flw_FLOW_CONTROL_LITERALS_Test__Flag6_06D8931												
-						flag-true	RAN_flw_FLOW_CONTROL_LITERALS_Test__Flag6_06D8931																					
-						Test	read0	read0	50130					Fail									flag-true	RAN_flw_FLOW_CONTROL_LITERALS_Test__Flag6_06D8931				
-						Test	read0	read0	50140					Fail									flag-true	RAN_flw_FLOW_CONTROL_LITERALS_Test__Flag6_06D8931				
+						defaults										r_flw_FLOW_CONTROL_LITERALS_Test__Flag6_06D8931												
+						flag-true	r_flw_FLOW_CONTROL_LITERALS_Test__Flag6_06D8931																					
+						Test	read0	read0	50130					Fail									flag-true	r_flw_FLOW_CONTROL_LITERALS_Test__Flag6_06D8931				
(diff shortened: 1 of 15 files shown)


redxeth:
Any comments?  Should I just merge it?  Need to make a new release.  @ginty @priyavadan @pderouen @chrisnappi

pderouen:
I've been holding my peace on this one. The changes suit our purposes. Thank you for this PR. From what I can tell the change maintains the previous behavior as the default. So, I think it's good to go. @priyavadan, feel free to weigh in.

The only update I would request is to document these new options in the guides page.

info-rchitect:
> I've been holding my peace on this one. The changes suit our purposes. Thank you for this PR. From what I can tell the change maintains the previous behavior as the default. So, I think it's good to go. @priyavadan, feel free to weigh in.
> 
> The only update I would request is to document these new options in the guides page.

Agreed on the docs update

redxeth:
Yep-- I was expecting the doc thing from you guys.   Can someone remind me how to update the docs again for something OrigenTesters specific?  Is there a procedure for this somewhere?

First update docs in the plug-in-- then?  Will Origen main guides pick up the change automatically?

redxeth:
> It's been a while. But, I believe you update the guides in `templates/origen_guides/program ` of origen_testers. After it's merged, the website can be updated by running `web compile -r` from origen.

Yes `web compile -r` is to regen and publish a new website-- I forgot about the way they are shared-- so the web publish / deploy within the OrigenTesters plug-in will only update that folder on the main origen website looks like.

redxeth:
Ok so in the process of making some updates, I'm seeing lots of lint changes-- I assume that's what all these recent rubocop PRs are about? 

So to confirm @priyavadan @pderouen @CodyBriscoe  -- my options are either to confirm the rubocop recommendations don't break OrigenTesters or add the exceptions? 

And all of this because of the rubocop upgrade needed for Ruby 3.1 ?   

UGH-- this is why people are so loathe to make Origen updates on their own (I'm doing this for someone else)

redxeth:
If someone could help fix would appreciate it-- having to go look through multiple PRs-- If this breaks many plugins then we have a lot of 'splaining to do lucy'

redxeth:
> Since it's my group causing the need for this PR and me that made the Ruby 3 updates, I should probably help figure out the right solution.

Thanks @pderouen !   We should as a group decide on how to handle 'breaking' changes done in Origen and all plug-ins.   If there is a breaking change (and here I realize the code isn't broken but needs to be upgraded to match some upgrade in a plug-in). 

Do poeple still use the discord server?   I checked and see the latest post from last year.

Perhaps github issues?  If something new we think may break others then file an issue with attention to the core team, etc., of what things need to be done.  Or add a deprecate warning somehow.  

Or if it has already broken something, then file an issue here - which I can do and add you @pderouen so that in case others encounter this they understand it's a known issue and how to resolve.

pderouen:
> > Since it's my group causing the need for this PR and me that made the Ruby 3 updates, I should probably help figure out the right solution.
> 
> Thanks @pderouen ! We should as a group decide on how to handle 'breaking' changes done in Origen and all plug-ins. If there is a breaking change (and here I realize the code isn't broken but needs to be upgraded to match some upgrade in a plug-in).
> 
> Do poeple still use the discord server? I checked and see the latest post from last year.
> 
> Perhaps github issues? If something new we think may break others then file an issue with attention to the core team, etc., of what things need to be done. Or add a deprecate warning somehow.
> 
> Or if it has already broken something, then file an issue here - which I can do and add you @pderouen so that in case others encounter this they understand it's a known issue and how to resolve.

You were on all of the issues and PR's for the Rubocop update. All of the discussions about it were had in the monthly global team calls. Agreed, there could also be a bulletin or something of that sort. Unfortunately, there just wasn't a way to stay current with Ruby and keep the ancient version of Rubocop that origen was previously locked with. We'll see if the ruleset can be further refined to make it less disruptive.

By the way there is still 1 open PR on origen that is needed for it to work with Ruby > 3.0.

pderouen:
I am working through the lint errors and will push an update to this PR soon. Some of the failures seem to be highlighting bugs. For those, I will just disable the check for that piece of code and file an issue rather than piling on to this PR.

redxeth:
> You were on all of the issues and PR's for the Rubocop update. All of the discussions about it were had in the monthly global team calls. 

Yeah sorry-- have a hard time keeping up sometimes with everything.

redxeth:
> I am working through the lint errors and will push an update to this PR soon. Some of the failures seem to be highlighting bugs. For those, I will just disable the check for that piece of code and file an issue rather than piling on to this PR.

Wow thanks for your help!

pderouen:
> > I am working through the lint errors and will push an update to this PR soon. Some of the failures seem to be highlighting bugs. For those, I will just disable the check for that piece of code and file an issue rather than piling on to this PR.
> 
> Wow thanks for your help!

It's the least I could do. The lint fix is coming in just a minute. I apologize in advance for the huge number of additional diffs. But, since we already reviewed the content of the PR and approved it hopefully won't be an issue.

pderouen (approved):

PR #199, opened 2023-08-30
Multi profile charz

CodyBriscoe:
**_Sidenote_** just calling out a side bug that I fixed that wasn't the main point of this PR, is that the auto ID setting would fail on SMT7 since the variable names were too long. What used to be:
`options[:id] = "#{parent_test_name}_charz_#{charz_session.name}".to_sym`
is now:
`options[:id] = "auto_charz_id_#{md5_id}".to_sym`
where md5_id is a hash of the parent test name, the options hash, and the current charz session id.

---

For background, the charz API was originally written to only handle a single "charz instance" at any given time, and the thought process was if you wanted some extra routines ran or some other behavior, you would just modify the profile settings to do that.

We are now setting up an automated system to handle some standardized characterization, which is coming in and auto creating and setting charz profiles. The problem here is that if an engineer already had some charz they'd defined, it would be overwritten (pushed down the charz stack) by the new automated profile, and then restored after the test of interested had been processed.

This PR allows for multiple profiles to now be "appeneded" to the current session, rather than overwriting the current session via some new API calls: `charz_on_append` and its counterpart `charz_off_truncate`. These effectively make the charz stack a 2D array, a stack of stacks. For example:

```ruby
 # assume the profiles have matching routines defined and set
charz_on :profileA
func :my_test1
charz_on_append :profileB
func :my_test2
charz_off_truncate
func :my_test3
charz_off
```
This would generate these tests in the following order:
- my_test1
- my_test1_routineA
- my_test2
- my_test2_routineA
- my_test2_routineB
- my_test3
- my_test3_routineA


info-rchitect (approved):
LGTM, nice feature

PR #201, opened 2024-02-09
Stil groups

pderouen:
Previously if a pin group was listed in the pin_pattern_order statement it was treated as though it was a single pin in the signal definition and timing definition sections of the output STIL. This PR is to include all of the pins from the group (see the 1 approved output change for an example).

### DIFF
--- a/.github/workflows/regression_tests.yml
+++ b/.github/workflows/regression_tests.yml
@@ -18,7 +18,7 @@ jobs:
     - name: Update Env
       run: echo "ACTIONS_ALLOW_UNSECURE_COMMANDS=true" >> $GITHUB_ENV
     - name: Install Bundler
-      run: gem install bundler -v '> 2'
+      run: gem install bundler -v '2.1.4'
     - name: Install dependencies
       run: bundle install  
     - name: Gem Install Origen 
--- a/approved/stil/v93k_workout.stil
+++ b/approved/stil/v93k_workout.stil
@@ -1,6 +1,6 @@
 // ***************************************************************************
 // GENERATED:
-//   Time:    09-Feb-2024 13:10PM
+//   Time:    13-Feb-2024 10:23AM
 //   By:      Paul DeRouen
 //   Mode:    debug
 //   Command: origen g v93k_workout -t legacy.rb -e stil.rb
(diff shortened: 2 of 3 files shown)


info-rchitect:
> Previously if a pin group was listed in the pin_pattern_order statement it was treated as though it was a single pin in the signal definition and timing definition sections of the output STIL. This PR is to include all of the pins from the group (see the 1 approved output change for an example).

Is this supposed to be optional?

pderouen:
> > Previously if a pin group was listed in the pin_pattern_order statement it was treated as though it was a single pin in the signal definition and timing definition sections of the output STIL. This PR is to include all of the pins from the group (see the 1 approved output change for an example).
> 
> Is this supposed to be optional?

I wasn't planning to make it optional. The previous output fails to compile. If it needs to be optional, I can do that.

coreyeng (commented):
I know very little about STIL, so I'll abstain for those that are more familiar to review.

pderouen:
> I know very little about STIL, so I'll abstain for those that are more familiar to review.

I know very little about STIL as well. I only tried STIL output for the first time a few days ago and discovered I have to list each pin individually in the pin_pattern_order statement to get a working STIL output.

pderouen:
@info-rchitect let me know if you need it to be optional. I was working from the assumption that this was an oversight/bug.

info-rchitect (approved):
@pderouen OK looks goot to me.

pderouen:
Thanks @info-rchitect. I will wait another few days in case anybody else wants to weigh in.

I was able to get my hands on the STIL standard and skimmed through it. It looks like there are a few other ways to resolve this issue (define as a signal bus, or define the porta & b groups after defining each signal individually). I prefer the way I have it implemented in this PR since we're lumping all of the signals together in 1 big group anyway when creating vector data. It also avoids complications when a pin group has a mix of waveforms for constituent pins (which generates an error from Origen when building the wave definitions).

arkohler:
As you mentioned, you can do a bus syntax
Signals{
  porta[7..0] InOut;
}
Otherwise, you'll need to define signals individually within the Signals{} block like you have it currently.
I do like the idea of auto-creating groups for these pin lists, and then you can use them or use the ALL group in the pattern.  I personally think that the Pattern block should contain the minimum group list to describe the pin changes, but I think that may be out of scope for this PR.

Bonus points if you want to go back and add the pin group definitions, otherwise this looks fine to me.

pderouen:
> Bonus points if you want to go back and add the pin group definitions, otherwise this looks fine to me.

Bonus points!!!??  How could I turn down bonus points! :)  I don't think it would be difficult to define the groups.

pderouen:
@arkohler, is that change what you had in mind?

arkohler:
Yes @pderouen, this looks good to me.

PR #202, opened 2024-02-10
remove extra spaces before converting to symbol

pderouen:
The reverse compiler (from a binary pattern file back to an ASCII pattern file) adds extra spaces to the pin list. The origen decompiler was turning that into a list of pins with a lot of spaces in their names (which don't exist in the app).


chrisnappi (approved):
LGTM

PR #203, opened 2024-04-05
Stil update

pderouen:
* Bug fix for undefined Waveset0 that was getting printed in the pattern output
* Added "use_timing_equations" option to define timing as equations
* Added callback method to set period size when converting an existing pattern to STIL


pderouen:
Suppose I should add at least 1 test for the new timing equation feature...

pderouen:
> I had to look at ruby 2.6 regressions recently as well for an upcoming PR.
> 
> I also disabled it on my branch too, however digging a bit more it seems like origen depends on installing rubocop-ast, and the version it needs fails to install on ruby 2.6. I'll look a bit more next week but if you happen to have some thoughts, I would be interested in them.

Do we still have any 2.6 users? I don't see much value in keeping the 2.6 checks. I'd rather spend effort adding 3.x checks and migrating to 3.x.

It may be as simple as adding a gem install for the right rubocop-ast version with the 2.6 workflow. I need to add a test or 2 for the new features I added anyway. I can take a swing at fixing 2.6 when I push the new test(s).

pderouen:
@priyavadan it was just that 1 issue installing origen to ruby 2.6. After updating the workflow to first install a specific version of rubocop-ast before installing origen, the 2.6 checks can run again.

Interested in any feedback on my code changes.

pderouen:
Thanks @priyavadan. I'll wait a few days before merging in case anybody else wants to weigh in.

PR #208, opened 2024-05-21
SMT8 does not require an initialization of the loop variable, it erro…

JamesCreamer:
SMT8 errors if the loop variable is initialized like other flow variables. The variable population had to be removed for it to run properly on the tester.

### DIFF
--- a/approved/v93k_flowgrouping/OrigenTesters/flows/FLOW_CONTROL.flow
+++ b/approved/v93k_flowgrouping/OrigenTesters/flows/FLOW_CONTROL.flow
@@ -9,10 +9,6 @@ flow FLOW_CONTROL {
     in EXTRAS = -1;
     in FLAG1 = -1;
     in JOB = "";
-    in LOOP_VARIABLE = -1;
-    in LOOP_VARIABLE1 = -1;
-    in LOOP_VARIABLE2 = -1;
-    in LOOP_VARIABLE3 = -1;
     in LT_VARIABLE = -1;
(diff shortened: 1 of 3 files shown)


pderouen:
It figures that the regression checks would be broken again. It _has_ been more than a month since the last time it needed a work around: https://github.com/Origen-SDK/origen_testers/commit/cdb01959710b32d6eb72ace00ec8ddaf1da6907d

I guess it was due.

pderouen:
> Why is windows 2.7 failing with this failure? https://stackoverflow.com/questions/64300670/ruby-gem-byebug-5-0-0-failing-on-install-due-to-make-error
> 
> Why is windows 2.6 build quitting out of the dependency build? I don't see anything that would cause the bundler evaluation to quit.

As soon as any build fails all pending builds stop. I'm not sure why byebug install would suddenly stop working. But, it seems to pop up on various gems during the GA builds. You might try removing the byebug version dependency in the Gemfile, or fix it some other version.

ThaoHuynhFsl:
LGTM.  Picked up this branch, generate packages on my current active project, ran it offline and online, it works fine.

ThaoHuynhFsl:
> Looks good to me.
> 
> I am bit surprised by this SMT8 behavior, though it does make sense.
> 
> @ThaoHuynhFsl please review?

looks like the initialization is inherit from out variable which required the initialization.

ThaoHuynhFsl (approved):

PR #209, opened 2024-06-12
Allow Ruby loop keyword within flow

rlaj:
Revert to Ruby loop keyword functionality if only block is given with no args, otherwise treat as testflow loop api.


pderouen:
Looks good to me.

pderouen (approved):

JamesCreamer (approved):
Looks good. The test case perfectly covers the observed issue

ThaoHuynhFsl (approved):
NXP test program generated fine with no issues or changes

PR #210, opened 2024-06-14
add option to add insertion in the flow path

ThaoHuynhFsl:
This is an update to the commit on 10/6/2023.  We need to keep the existing behavior that is no insertion in the flow path.

Add in an option to add the insertion.  In the environment/v93k_smt8.rb

add insertion_in_the_flow_path: true


JamesCreamer (approved):
This change works as intended on my end.

PR #212, opened 2024-06-18
Force integer and double to their respective types

JamesCreamer:
SMT8 requires the types to be maintained for the parameters. Without forcing the type in the parameters, providing a string to integers and doubles will show up as strings in the final output.


ThaoHuynhFsl:
Regressed the updates and I like the changes.  Ran offline on my active projects with no issue.

ThaoHuynhFsl (approved):

rlaj (approved):
LGTM

PR #215, opened 2024-06-28
Fix the period alias on nested parameters

JamesCreamer:
The alias handling for nested parameters was not matching the other core locations. I had to update here to align with expected key handling.


rlaj (approved):
Looks good.

ThaoHuynhFsl (approved):
I am good with the changes.  Currently we dont use nested parameters.

PR #217, opened 2024-08-19
Convert updates

pderouen:
Various bug fixes for issues uncovered while running origen convert on ascii patterns from various sources:
    1) End of line comments that origen creates when generating with -v option cause the vector to be detected as a label
    2) Trailing ":S" on pin names needs to be removed (ex: JTAG:S, CLKS:S, etc.)
    3) Blank lines cause a run time error


rlaj (approved):
LGTM.  Didn't break anything here.

pderouen:
> LGTM. Didn't break anything here.

Thanks!

PR #220, opened 2024-11-05
Feature/smt8 ods advantest prefer way

ThaoHuynhFsl:
Currently software bin description is using test suite name.  Allow Software Bin description to be passing from app.

Switch the column orders 
Software Bin Description and Software bin - Software Bin Description is the column before Software bin
Hardware Bin Description and Hardware bin - Hardware Bin Description is the column before Hardware bin

This only impacts the .ods file


ThaoHuynhFsl:
> LGTM! Thank you for the updates.

thanks

PR #221, opened 2024-11-05
eof charz generation has a valid usecase for referencing the stored instance - stop the spam

CodyBriscoe:
Minor changes to prevent spamming the console with deprecation warnings when using end of flow charz generation placement, which has a valid use case for using the stored instance. Even outside eof, referencing charz_session.<attr> via charz_instance.<attr> is always valid.


ThaoHuynhFsl:
ran fine with our apps.  thanks

PR #222, opened 2024-11-08
Skip typecast for V93K parameter when default value is empty string

rlaj:
Prevent empty default value from being converted to default type value.  Ex. `''.to_i => 0`


ThaoHuynhFsl (approved):
LGTM.  Ran fine with our apps

PR #235, opened 2026-02-09
Feature/ruby 4 0

priyavadan:
this has a lot of files that changed but most are "lint" fixes.  Please do rely on the  CI  results confirming that functionality is retained. 

expected that this will fail on Ruby 4.0 until  `origen` is released.


pderouen (approved):
Seems pretty straight forward. Thanks for this update.

